Issue #37, opened 2020-03-23
exr::Error should have a default formatter and impl std::error::Error


mandeep:
`exr::Error` should have a default formatter so that `eprintln!("{}, error");` will work.

Also, `exr::Error` should impl `std::error::Error` that way you can return from a function as `Result<(), Box<dyn Error>`. 

If you would like help with this let me know and I will submit a PR.

johannesvollmer:
Thanks for the issue. Sounds like a good idea. Can you nevertheless please show me some example usage of the new features you suggested? Why exactly is it important to implement them?

If you are curious, try to implement it yourself. The error type is defined inside `src/error.rs`. If any change outside of that Module is necessary, please let me know before you start working.

mandeep:
The main reason is for error propagation. When you have a `main()` function that can error out in many ways, it helps when all of the functions that return `Result` do so with `std::error::Error` implemented for the `Err`. That way we can write a function prototype like `fn main() -> Result<(), Box<dyn Error>` and not have to worry about the different types of errors that different libraries return.

johannesvollmer:
Ok, perfect. If you want to, do it. Also, feel free to ask anything here, especially if there's something that lacks documentation or otherwise needs more explanation in the library :)

johannesvollmer:
fixed by #39

PR #39, opened 2020-03-26
Add impl of std::error::Error and default formatter

mandeep:
fixes #39 

This PR adds an impl for std::error::Errror and the default formatter.

Only the IoError is handled since the other are types `Cow<'static, str>` (I'm not sure why they aren't just String). Let me know what you think. I am not able to test it as the changes on the master branch break a lot of things from 0.7.3.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -6,6 +6,8 @@ use std::io::ErrorKind;
 pub use std::io::Error as IoError;
 pub use std::io::Result as IoResult;
 use std::convert::TryFrom;
+use std::error;
+use std::fmt;
 
 // Export types
 
@@ -67,6 +69,20 @@ impl From<IoError> for Error {
     }
 }
 
+impl error::Error for Error {
+    fn source(&self) -> Option<&(dyn error::Error + 'static)> {
+        match *self {
+            Error::Io(ref err) => Some(err),
+            _                  => None,
+        }
+    }
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", <Self as error::Error>::description(self))
+    }
+}
 
 /// Return error on invalid range.
 #[inline]
--- a/src/error.rs
+++ b/src/error.rs
@@ -80,7 +80,12 @@ impl error::Error for Error {
 
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{}", self)
+        match self {
+            Error::Io(err) => err.fmt(f),
+            Error::NotSupported(message) => write!(f, "not supported: {}", message),
+            Error::Invalid(message) => write!(f, "invalid: {}", message),
+            _ => write!(f, "{}", self),
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


johannesvollmer:
The error being a Cow instead of a plain String is an unnecessary optimization. We can replace that with a simple String later.

Before that though, we should make it possible to compile the project and run the tests. What exactly seems to be the problem there? The error module should not have changed that much, does pulling the changes from origin/master and then runing `cargo test --no-fail-fast` work?

mandeep:
`cargo test --no-fail-fast` works fine. I was just talking about the my application that would have tested these changes no longer runs due to the changes in the master branch.

johannesvollmer:
Oh ok, perfect! If it compiles, we can be confident. Let me check it out and then I will merge.

johannesvollmer:
The display implementation however, would probably match on the error type and write the Cow value to the formatter, instead of only handing IO errors. Would you try that?

mandeep:
How do you propose we convert from Cow<'static, str> to std::error::Error?

johannesvollmer:
The error implementation itself looks good. The display implementation however, what text exactly does it return for non-io errors?

mandeep:
I think Display would return the &str, however I haven't been able to test it. Do you have a simple application made with all the changes you made in master?

Another approach would be use a crate like quick-error to handle the creation of errors: https://crates.io/crates/quick-error

johannesvollmer:
>Do you have a simple application made with all the changes you made in master?

Yes, the examples are always up-to-date. Instead of reading into an array of pixels, you implement an interface for your own image type.
https://github.com/johannesvollmer/exrs/blob/1b158ccfe1744d0a04e41315db95a32ed7556d3e/examples/rgba.rs#L21-L54

This is more code than before, but avoids having to convert a strange array image to your own image type. If this does not explain it well enough, let me know, then I'm going to write some detailed documentation for the image module.

johannesvollmer:
What I meant is something like the following: 
```rust
    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
        match self {
              Io (error) => error.fmt(formatter),
              Unsupported (message) => write!("not supported: {}", message),
              Invalid (message) => write!("invalid: {}", message),
        }
    }
```

This would complete your PR in my eyes, improving the library a lot with low effort. Using a crate like `quick-error` would certainly improve the library, but require far more changes. I think it would be the best approach to finish this PR with detailed formatting messages, merge it, and then consider a new PR for bigger changes to the error handling. What do you think?

mandeep:
Yep makes perfect sense. I've gone ahead and pushed the necessary changes.

johannesvollmer:
Perfect! Thanks for your time and effort.

johannesvollmer:
Have you tried using the new rgba module? Was it straightforward?

johannesvollmer:
fixes #37

mandeep:
There were a lot of breaking changes so it will take time for me to upgrade. I liked how things were setup with the Pixels enum in 0.7.3. The GetPixels and SetPixels traits don't make sense to me. Instead of traits I think these could just be methods that are called.

johannesvollmer:
Thanks for the feedback! Some of it can be functions, but not all of it, yet. Have a look at the `generate_rgba` example; it contains an example on how to use a reader function instead. The writing however does not yet fully support functions yet. I hope that I can find a nice API design for that later. :)

PR #63, opened 2020-05-14
Fix tests

johannesvollmer:
Tests did not panic for invalid files. 

Now they only allow `NotSupported` but not `Invalid` errors.


johannesvollmer:
Also, add API for `Text.chars()`

PR #64, opened 2020-05-14
Proper fix for ZLib decompression.

dgsantana:
I added the previous fix related to compression.
Also added the error conversion from string to avoid adding too much code just to map the error on the zip.rs.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -69,6 +69,13 @@ impl From<IoError> for Error {
     }
 }
 
+/// Special case for `inflate_bytes` where the error is a String
+impl From<String> for Error {
+    fn from(msg: String) -> Self {
+        Error::Invalid(msg.into())
+    }
+}
+
 impl error::Error for Error {
     fn source(&self) -> Option<&(dyn error::Error + 'static)> {
         match *self {
--- a/src/error.rs
+++ b/src/error.rs
@@ -69,13 +69,6 @@ impl From<IoError> for Error {
     }
 }
 
-/// Special case for `inflate_bytes` where the error is a String
-impl From<String> for Error {
-    fn from(msg: String) -> Self {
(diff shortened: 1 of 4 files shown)


dgsantana:
All local tests passed ;).

dgsantana:
Local tests didn't gave unexpected errors :).

src/error.rs
```diff
@@ -69,6 +69,13 @@ impl From<IoError> for Error {
     }
 }
 
+/// Special case for `inflate_bytes` where the error is a String
+impl From<String> for Error {
+    fn from(msg: String) -> Self {
+        Error::Invalid(msg.into())
+    }
+}
+
```
johannesvollmer, line 78: I think, in the future, someone could be surprised to find out that all Strings are implicitly converted to `Invalid`, but never to `Unsupported` or the like.  I suppose it would be more "obvious" to avoid this implementation and instead use `compressor.finish().map_err(|string| Error::invalid(string))` inside `zip.rs`. What do you think?

dgsantana: I have to agree, this can bring more trouble in the future than it should.

johannesvollmer:
Ok, wonderful! Just waiting for those checks and then we can merge.

johannesvollmer:
Thanks for your patience and the time you spent to improve exrs! 💯

dgsantana:
No problem, glad I could help. Next "problem" is supporting the rest of the
compressions. I already saw your piz branch. I will try to give a hand on
the conversion.

Johannes Vollmer <notifications@github.com> escreveu no dia quinta,
14/05/2020 à(s) 16:13:

johannesvollmer:
Yes, getting the compression to work is the number one priority right now, have fun :D

PR #76, opened 2020-05-28
Huffman decode fix

dgsantana:
Left your code for testing in there, just fixed the problem.

### DIFF
--- a/src/compression/piz/huffman.rs
+++ b/src/compression/piz/huffman.rs
@@ -32,10 +32,10 @@ const ENCODING_TABLE_SIZE: usize = (1 << ENCODE_BITS) + 1;
 const DECODING_TABLE_SIZE: usize = 1 << DECODE_BITS;
 const DECODE_MASK: usize = DECODING_TABLE_SIZE - 1;
 
-const SHORT_ZEROCODE_RUN: i64 = 59;
-const LONG_ZEROCODE_RUN: i64 = 63;
-const SHORTEST_LONG_RUN: i64 = 2 + LONG_ZEROCODE_RUN - SHORT_ZEROCODE_RUN;
-const LONGEST_LONG_RUN: i64 = 255 + SHORTEST_LONG_RUN;
+const SHORT_ZEROCODE_RUN: u64 = 59;
+const LONG_ZEROCODE_RUN: u64 = 63;
+const SHORTEST_LONG_RUN: u64 = 2 + LONG_ZEROCODE_RUN - SHORT_ZEROCODE_RUN;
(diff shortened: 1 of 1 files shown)


dgsantana:
Ok found that one of your comment uncompressed arrays causes an error, now in different part.

johannesvollmer:
Awesome. I currently don't have the time, to merge this, but I will in a few days. Looks good so far!

johannesvollmer:
Actually, as this does not merge into master but the PIZ branch, let me just do it via the github website ^^

dgsantana:
Cool. This branch needs some more cleanup on println! and inspect!. I may have time to rustify a bit later today.

johannesvollmer:
Absolutely! piz.rs has a lot of stuff going on that was only there to debug. Furthermore I'll do the following:
- check 'as' casts if they should return errors
- do not crate large arrays (vec should not be slower most of the time)
- no unwrap
- no assert triggered by file contents
- simplify C patterns by using Rust std lib
- compute exactly required vector sizes (if possible)
- remove comments that contain C code
- return instead of mut parameter
- iter().collect() instead of push
- use functions that to reuse code (sometimes the C code could have done that too)

johannesvollmer:
I find it hard but often rewarding to try to understand the bigger picture, and try to get rid of the stuff they did because C and not because the algorithm required it

johannesvollmer:
I have made some basic cleanup commits (also in huffman, but only really simple changes there!).

When I ran the tests, three images fail because something is wrong with the "native" piz compression. The images are 
- "tests/images/valid/openexr/IlmfmlmflmTest/test_native2.exr"
- "tests/images/valid/openexr/IlmfmlmflmTest/test_native1.exr"
- "tests/images/valid/openexr/TestImages/AllHalfValues.exr"

I think this problem when an image has only half channels. Something is not right with the "native" compression, but I was not able to figure it out in the short time I had. I'll have to figure that out later

johannesvollmer:
https://github.com/johannesvollmer/exrs/blob/2ae20b4823f572487fa7b8a51153a4cc3ebc1d1d/src/compression/piz/mod.rs#L124-L130 
and 
https://github.com/johannesvollmer/exrs/blob/2ae20b4823f572487fa7b8a51153a4cc3ebc1d1d/src/compression/piz/mod.rs#L191-L197

johannesvollmer:
Just ignore the tests failing for these three images for now, until I find the time to fix that :)

johannesvollmer:
(I removed the inspect macro, because I found out Rust has it already, it's called `dbg!`)

PR #85, opened 2020-07-19
add more attributes and auto crop

### LINKED ISSUE #83: Add support for missing common attributes
Specifically I miss `near`, `far`, `fieldOfView*` & `software` from below list (which is a common EXR from a professional 3D production offline renderer). 

```
exrheader out_noise.exr

file out_noise.exr:

file format version: 2, flags 0x0
channels (type chlist):
    A, 16-bit floating-point, sampling 1 1
    B, 16-bit floating-point, sampling 1 1
    G, 16-bit floating-point, sampling 1 1
    R, 16-bit floating-point, sampling 1 1
compression (type compression): zip, multi-scanline blocks
dataWindow (type box2i): (0 0) - (2047 2047)
displayWindow (type box2i): (0 0) - (2047 2047)
far (type float): 1e+38
fieldOfViewHorizontal (type float): 30
fieldOfViewVertical (type float): 30
lineOrder (type lineOrder): increasing y
near (type float): 0.001
pixelAspectRatio (type float): 1
screenWindowCenter (type v2f): (0 0)
screenWindowWidth (type float): 1
software (type string): "3Delight 2.1.9 darwin-x86_64 (Jun 24 2020, 2a265d) "Re-Animator""
type (type string): "scanlineimage"
worldToCamera (type m44f):
   (1 0 0 0
    0 1 0 0
    0 0 1 0
    0 0 0 1)
worldToNDC (type m44f):
   (3.73205 0 0 0
    0 3.73205 0 0
    0 0 1 1
    0 0 -0.001 0)
```

### LINKED ISSUE #84: Add autocrop functionality
This would compute a tight(ened) data window by looking at all pixels in all channels. Empty pixels are those with value zero except in depth channels where it's those bigger or equal to the max value of the type (float or half).

This is a convenience function many EXR writers for 3D renderers offer. 

Could be a parameter to `infer_meta_data()` maybe? For most uses cases having such a shrunken data window is the desirable behavior so it could even be a default that has to be turned off explicitly.

johannesvollmer:
closes #84 and closes #83 (please reopen if desired)

- add some more attributes and auto-crop functionality
- also rename existing stuff, for example IntRect and some layer attributes

### DIFF
--- a/src/meta/header.rs
+++ b/src/meta/header.rs
@@ -34,8 +34,8 @@ pub struct Header {
     /// In what order the tiles of this header occur in the file.
     pub line_order: LineOrder,
 
-    /// The resolution of this layer. Equals the size of the data window.
-    pub data_size: Vec2<usize>,
+    /// The resolution of this layer. Equivalent to the size of the `DataWindow`.
+    pub layer_size: Vec2<usize>,
 
     /// Whether this layer contains deep data.
     pub deep: bool,
@@ -96,10 +96,10 @@ pub struct ImageAttributes {
     /// The time code of the image.
     pub time_code: Option<TimeCode>,
 
-    /// Optional attributes. Contains custom attributes.
+    /// Contains custom attributes.
     /// Does not contain the attributes already present in the `ImageAttributes`.
     /// Contains only attributes that are standardized to be the same for all headers: chromaticities and time codes.
-    pub custom: HashMap<Text, AttributeValue>,
+    pub other: HashMap<Text, AttributeValue>,
 }
 
 /// Does not include the attributes required for reading the file contents.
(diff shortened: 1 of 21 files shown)


src/meta/header.rs
```diff
@@ -257,6 +257,28 @@ impl LayerAttributes {
     pub fn with_position(self, data_position: Vec2<i32>) -> Self {
         Self { layer_position: data_position, ..self }
     }
+
+    /// Set all common camera projection attributes at once.
+    pub fn with_camera_frustum(
```
virtualritz, line 262: Nice!

johannesvollmer: Thanks, that's the best I could do without implementing a bunch of matrix operations that would be used nowhere else :)

johannesvollmer: Can be used like this: `with_camera_frustum(mat1, mat2, (fov_x, fov_y), near..far)`

PR #89, opened 2020-07-25
Deduplicate Image API

johannesvollmer:


johannesvollmer:
Closing this because it is contaminated with git LFS which would require a monthly fee. I'll make a new PR for deduplicating the API without LFS

PR #99, opened 2021-02-02
extend rgba to all pixels

johannesvollmer:
Introduce type `SpecificChannels`


johannesvollmer:
Two tests still fail (write 2 and 7 channels)

PR #106, opened 2021-04-11
Implement Recursive code generation with macros, should fix #102

### LINKED ISSUE #102: Writing images with more than 8 channels
First of all thanks for making this crate, it's very useful and pleasant to use.

When the example [`4_specific_write.rs`](https://github.com/johannesvollmer/exrs/blob/c4a591f4b21580f5d4478455e7a28416ef770e80/examples/4_specific_write.rs) is changed so that the image contains more than 8 channels the code no longer compiles and instead results in the following error:

> the trait `IntoRecursive` is not implemented for `(exr::prelude::f16, f32, u32, f32, f32, f32, f32, f32, f32)`

This makes sense, indeed `IntoResursive` is only [implemented](https://github.com/johannesvollmer/exrs/blob/c4a591f4b21580f5d4478455e7a28416ef770e80/src/image/recursive.rs#L93-L101) for tuples up to size 8. In that file there's a comment saying

> // TODO use a macro to generate these impls!

Also, there's what looks to be an existing macro implementation at the bottom of the file. What was the blocker there? Was there an issue with the macro? Some other reason to not continue implementing this? If it's simply a time issue I'm willing to give it a shot and submit a PR.

In the meantime, does this crate currently have an existing workaround for this? In general, is it possible to write images with a number of channels or even a chanel data type not known at compile time?

KarelPeeters:


KarelPeeters:
I learned a lot about macros in Rust while writing this 🙂. [The macro metavar RFC](https://github.com/markbt/rfcs/blob/macro_metavar_expr/text/0000-macro-metavar-expr.md) would have come in handy!

The current implementation has a big issue where the default recursion_limit of 128 is exceeded for tuple larger than size 12. To get the final macro invocation to look neat like this:

```rust
generate_all!(
        A, B, C, D, E, F, G, H;
        0, 1, 2, 3, 4, 5, 6, 7;
); 
```

the input sequences basically need to be reversed twice because of macro limitaions. This uses up a lot of the recursion limit in an `O(n^2)` manner so it doesn't really work that well for large tuples. I assume compilation will eventually start to slow down because of this as well. We could just increase the limit for this crate with something like `#![recursion_limit="1024"]` but even that would only get us to tuples of size 41 (to give some perspective).

I could drop the neatness and throw out a bunch of ulgy macro code (basically all of `generate_all` and `generate_all_reversed` and do something like this instead:

```rust
generate_single!(A; A; 0);
generate_single!(A, B; B, A; 1, 0);
generate_single!(A, B, C; C, B, A; 2, 1, 0);
...
```

Is that still an okay solution? It should at least be a lot simpler and faster!

johannesvollmer:
Awesome! The verbose version you proposed is already better than what we have now. I'd like to use that :)

johannesvollmer:
Will it be possible to declare the reversed order, but still only have one macro invocation? Like:

```rust
generate_impls! {  A,B,C; B,C,A; 1,2,3; 3,2,1  }
```

At each recursive step of the macro, emit one impl?

KarelPeeters:
No unfortunately that still doesn't solve the issue, when recursing you can only cut off the first element in a sequence, not the last one. A small example:

```rust
macro_rules! demo {
    (first; $first:literal $(,$rest:literal)*) => { };
    (last; $($rest:literal,)* $last:literal) => { };
}
```

Now `demo!(first; 1, 2, 3)` works but `demo!(last; 1, 2, 3)` doesn't, it's complaining that parsing is ambigious.

I'll commit the semi-manual solution in a few minutes, then I think this PR is ready.

One thing I was confused about is why `NoneMore` implements `ToRecursive`, it's already recursive right? But when I removed that impl the rest of the crate doesn't compile anymore. I assume that could be fixed but I guess that's not backwards compatible.

KarelPeeters:
Okay so I went until tuple size 32, but I guess either 16 or 64  (or 26 😉) work just as well.

I generated the code with a small python script here: https://gist.github.com/KarelPeeters/1bd735d98d84cce7cdac8aeeb8970196

Let me know if you want me to add that to the repo too.

johannesvollmer:
> NoneMore implements ToRecursive, it's already recursive right

Just as any `T` implements `Into<T>`, this simplifies some special cases in the type system. It's not something to worry about.

> when recursing you can only cut off the first element in a sequence

Yeah, I think I didn't express myself clearly. We can't avoid having to manually type the sequence a second time in reverse. I just wondered whether it would be possible to call a single macro instead of calling the macro 32 times with increasing elements. I'll code an example when I'm back at the computer :)

KarelPeeters:
>Just as any T implements Into<T>, this simplifies some special cases in the type system. It's not something to worry about.

Ah I missed that `Recursive` also implements it, then it's consistent and does make sense.

>Yeah, I think I didn't express myself clearly. We can't avoid having to manually type the sequence a second time in reverse. I just wondered whether it would be possible to call a single macro instead of calling the macro 32 times with increasing elements. I'll code an example when I'm back at the computer :)

Hmm I'm interested to see if it's indeed possible, I think the fundamental problem is that it's easy to go from `D, C, B, A` to `C, B, A` it's impossible to go from the sequence `A, B, C, D` to `A, B, C` at the same time. Of course if you figure out a way that's great!

johannesvollmer:
> it's impossible to go from the sequence A, B, C, D to A, B, C at the same time

Ahaaaaaa yes, now I get it, sorry, your right! :D I didn't see that this would be required because we need to remove the last element from the non-reversed list.

> Okay so I went until tuple size 32, but I guess either 16 or 64 (or 26 😉) work just as well. 

Great. I think 32 could be enough. I support the Rust philosophy of "if you have more than 32 elements in a tuple, you should probably reconsider what you're doing". Also, one could always use the dynamic approach, just as you did until now.

> I generated the code with a small python script

That's ok. I'd rather not include the script in the repository though, because if we ever find that we need to increase the number, we should reconsider what we are doing (and I also hope Rust will get even better by then and will not require manually generating this source code at all).


## Merging
I'll have a look at the code. On first glance, it looks excellent.

johannesvollmer (approved):
Great documentation! :)

johannesvollmer:
I wonder why my local dev branch builds on my computer, even though it has the trailing semicolon. Works now, I guess, so we don't need to worry right :D

johannesvollmer:
The github tests don't seem to run, still _Waiting for status to be reported_

johannesvollmer:
I'll try closing and reopening this pull request

johannesvollmer:
Alright I'll run the tests locally...

johannesvollmer:
... Tests succeed on my machine.

PR #110, opened 2021-04-15
remove unnecessary semicolons

Narann:
Those semicolons seems to be useless and some of them prevents compilation (rust 1.51).


johannesvollmer:
Very strange. In the previous pull request I already had the same issue one, I thought it was resolved by now. Do you have an idea how I can get my compiler to recognize these errors?

johannesvollmer:
I want to fix the github checks, I need this when I'm on the computer

Narann:
> Do you have an idea how I can get my compiler to recognize these errors?

TBH, I really don't know why some are just warning and only one or two block compilation.

If you use PyCharm, "`rustfmt` on save" actually remove those semicolons.

PR #111, opened 2021-04-16
declare Requirements fields public

Narann:
Command line tool will need those for parsing headers.

### DIFF
--- a/src/meta/mod.rs
+++ b/src/meta/mod.rs
@@ -66,22 +66,22 @@ pub struct Requirements {
 
     /// This library supports reading version 1 and 2, and writing version 2.
     // TODO write version 1 for simple images
-    file_format_version: u8,
+    pub file_format_version: u8,
 
     /// If true, this image has tiled blocks and contains only a single layer.
     /// If false and not deep and not multilayer, this image is a single layer image with scan line blocks.
-    is_single_layer_and_tiled: bool,
+    pub is_single_layer_and_tiled: bool,
 
     // in c or bad c++ this might have been relevant (omg is he allowed to say that)
     /// Whether this file has strings with a length greater than 31.
     /// Strings can never be longer than 255.
-    has_long_names: bool,
+    pub has_long_names: bool,
 
     /// This image contains at least one layer with deep data.
-    has_deep_data: bool,
+    pub has_deep_data: bool,
 
     /// Whether this file contains multiple layers.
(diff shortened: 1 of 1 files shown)


johannesvollmer:
I see no problem with making the fields public.
 
But why exactly would this be interesting to anyone? The version is always `2` and the other attributes can be inferred from the real meta data, for example the number of layers and whether a layer is tiled. Which of the tools are you talking about and why does it need to do any parsing? :)

Narann:
> Which of the tools are you talking about and why does it need to do any parsing?

The `exrheader` I am writing. Official `exrheader` provides this information:

```
file format version: 2, flags 0x0
```

Those information are what are inside the first `int` of an EXR file, after the magic number.

I consider such tool should expose what are inside the file, regardless if it make sense to do so : If it's in the file, show it.

Another way to say it is: Show everything that are not pixels.

As `exr` store those information when parsing the file, I consider it would make sense to simply use them.

But maybe there are some under the hood reason to not mark field `pub`, maybe it have other valid considerations. As it's simply a `int` parsing, I can do it myself, no worry.

johannesvollmer:
> As it's simply a int parsing, I can do it myself

I'm not to make you do that :D As I said, I'm just interested, there is nothing wrong with making the fields public

> Show everything that are not pixels.

This already answers my question :)

Narann:
> But you are going to actually print the human readable information instead of only the encoded integer, right?

Yes ! Raw straight, for now, it looks like this:

```
file_name: foo.exr

file_format_version: TODO
flags: TODO
layer count: 1
layer #0
  compression: zip block compression
  block: scanline
  line_order: increasing
  layer_size (v2u): (2048, 2048)
  display_window (box2i): (0, 0) - (2048, 2048)
  pixel_aspect_ratio (f32): 1
  OTHER TODO
  layer_position (v2i): (0, 0)
  screen_window_center (v2f): (0, 0)
  screen_window_width (f32): 1
  channels:
    B, f16, sampling (1, 1)
    G, f16, sampling (1, 1)
    R, f16, sampling (1, 1)
```

johannesvollmer:
I mean, especially the `flags` field should actually be 3 separate fields (that happen to be crammed into a single integer when encoded, which is an implementation detail)

Narann:
Yes, the `flags: TODO` will be the list of flag:
```
flags:
  foo
  bar
```

PR #112, opened 2021-04-16
add b44 compression/decompression prototype

Narann:
Here is the first raw implementation of b44 compression. It technically works but need a huge amount of feedback to be nice.

I had to disable some asserts. I'm new to rust and am not comfy with some topics (memcpy).


johannesvollmer:
I just had a quick look. Nice progress. 

There's actually [two different B44 compression types](https://github.com/AcademySoftwareFoundation/openexr/blob/4c587457266cafb5d293d58963ec3c288fa4be2b/src/lib/OpenEXR/ImfCompressor.cpp#L138-L144), maybe we can add both of them in one go?

```cpp
    case B44_COMPRESSION: return new B44Compressor (hdr, maxScanLineSize, 32, false);
    case B44A_COMPRESSION: return new B44Compressor (hdr, maxScanLineSize, 32, true);
```

> [B44A] is a variation of B44, but with a better compression ratio for images with large uniform areas, such as in an alpha channel.

johannesvollmer:
Also, I have updated the tests to work with a threshold (equality check with a tolerance), [here](
https://github.com/johannesvollmer/exrs/compare/dev...prepare-lossless-compression-methods), and also activated the debug_assert for lossless compression methods.

If you want to check it out, you could merge that branch into this pull request, then we can check whether the new test design works for you.

Narann:
Wow, I didn't catch the `true/false` argument for compressor! Damn, I will look at the implication.

Thanks for the modification proposal, yes we will have to handle the lossless/lossy situation!

johannesvollmer:
> didn't catch the true/false argument

When I last had a look at your code, it was already there, but not connected to the B44A compression, it's the `opt[imize]_flat_fields` boolean :)

johannesvollmer:
Awesome!

src/compression/b44/mod.rs
```diff
@@ -0,0 +1,924 @@
...
+
+    if (s[0] & 0x8000) != 0 {
+        s[0] &= 0x7fff;
+    } else {
+        s[0] = !s[0];
+    }
+
+    for i in 0..16 {
+        s[i] = s[0];
+    }
```
Narann, line 246: Any idea on how to change this to a real rusty approach?

src/compression/b44/mod.rs
```diff
@@ -0,0 +1,924 @@
...
+}
+
+fn cpy_u16(src: &[u16], src_i: usize, dst: &mut [u16], dst_i: usize, n: usize) {
+    // assert_eq!(src.len(), dst.len());
+
+    // for i in 0..src.len() {
+    for i in 0..n {
+        dst[dst_i + i] = src[src_i + i];
+    }
+}
```
Narann, line 268: This code and the place using it are kinda ugly. It's supposed to be a `memcpy` and I'm not sure this code is the clearer to write.

src/compression/b44/mod.rs
```diff
@@ -0,0 +1,924 @@
...
+
+pub fn decompress(
+    channels: &ChannelList,
+    compressed: &ByteVec,
+    rectangle: IntegerBounds,
+    expected_byte_size: usize,
+    _pedantic: bool,
```
Narann, line 275: What `_pedantic` is supposed to do?

src/compression/b44/mod.rs
```diff
@@ -0,0 +1,924 @@
...
+            if remaining < byte_count {
+                // notEnoughData();
+                return Err(Error::invalid("not enough data"));
+            }
+
+            //memcpy (cd.start, inPtr, n);
+            for i in channel.tmp_start_index..(channel.tmp_start_index + byte_count) {
+                // TOTALLY WRONG
+                out[out_i + i] = compressed[in_i + i];
+            }
```
Narann, line 334: I need to review this part. It came from a moment when I didn't use a `tmp_u16_buffer`.

src/compression/b44/mod.rs
```diff
@@ -0,0 +1,924 @@
...
+                    //         .expect("write to in-memory failed");
+                    //     // out[(row0 + i)..(row0 + i + 1)] = s[i + 0].to_be_bytes();
+                    //     // out[(row1 + i)..(row1 + i + 1)] = s[i + 4].to_be_bytes();
+                    //     // out[(row2 + i)..(row2 + i + 1)] = s[i + 8].to_be_bytes();
+                    //     // out[(row3 + i)..(row3 + i + 1)] = s[i + 12].to_be_bytes();
+                    //     // out[row0 + i] = s[i + 0];
+                    //     // out[row1 + i] = s[i + 4];
+                    //     // out[row2 + i] = s[i + 8];
+                    //     // out[row3 + i] = s[i + 12];
+                    // }
```
Narann, line 454: I'm sure there is a better way to do this. Any advice is welcome. What could be the proper way to do `memcpy (row0, &s[ 0], n);` ?

src/compression/b44/mod.rs
```diff
@@ -0,0 +1,924 @@
...
+                    s[6] = tmp[row1 + 2];
+                    s[7] = tmp[row1 + 3];
+                    s[8] = tmp[row2];
+                    s[9] = tmp[row2 + 1];
+                    s[10] = tmp[row2 + 2];
+                    s[11] = tmp[row2 + 3];
+                    s[12] = tmp[row3];
+                    s[13] = tmp[row3 + 1];
+                    s[14] = tmp[row3 + 2];
+                    s[15] = tmp[row3 + 3];
```
Narann, line 739: Same `memcpy` problem here. I'm sure we can do better.

Narann (commented):
A whole cleanup is needed, but I would like to find a proper way to `memcpy` before.

src/compression/b44/mod.rs
johannesvollmer: Sadly, there is no convenience function for this in Rust `std`. 

Here is how I would do it:
```rust
let (first, rest) = s.split_first_mut().unwrap();  // split off first item, gain mutable access to both (only fails if slice would be empty)
for element in rest { *element = *first; } // feels like it should be a predefined function, but will always be optimized, no worries
```

Or maybe:
```rust
let first = s[0]; 
for element in &mut s[1..] { *element = first; }
```

1. it skips the first element when writing (which is not a big performance win, but delivers intent). 
2. In this PR in general, many index-based loops could be replaced with direct reference loops. This is easier to read and can be optimized by the compiler. So, I would always prefer `for x in &slice { }` over `for i in range { }`. You can use `for (elem, index) in slice.iter().enumerate() { }` if you also need the index for computing the element. This is also how the general Rust community prefers the loops :)

johannesvollmer:
In general, you could read the [slice docs](https://doc.rust-lang.org/std/primitive.slice.html). These methods can be applied to arrays, vecs, and slices of course. Take some time to get a feeling for what is possible with Rust `std` :)

src/compression/b44/mod.rs
johannesvollmer: Yes, seems like it is just a copy, but not from the beginning, but from some start index.

```rust
fn copy_u16(src: &[u16], src_i: usize, dst: &mut [u16], dst_i: usize, n: usize) {
    let source      = &    src[src_i .. src_i + n]; // make a slice from start index to end
    let destination = &mut dst[dst_i .. dst_i + n]; // make a mutable slice from start index to end

    destination.copy_from_slice(source); // copy all slice values

   // for educational purposes, here's how I would do that by hand, if the method above did not exist:
    for (src_elem, dst_elem) in source.iter().zip(destination.iter()) {  // iterate through both slices simultaneously
      *dst_elem = *src_elem; // copy each element
    }
}
```

Considering that the actual copying is a single function call, I would probably see if this function can be removed completely, and work with the slices directly. Maybe the whole scr_i and n variables are not necessary at all, if you work with slices all the time, instead of using indices. At least that's what I usually try to do.

src/compression/b44/mod.rs
johannesvollmer: This is a flag the user sets when reading or writing the image. If pedantic is true, then an error will be returned as soon as anything is missing in the file, or two values in the image contradict each other. If pedantic is false, then only fatal errors will be thrown. By default, reading an image is not pedantic, which means that slightly invalid files might still be readable. For example, if some attribute is missing but can be recomputed, this flag decides whether an error is thrown. Or if the pedantic flag is true and there are still bytes left after the decompression algorithm finished, an error is thrown, because this should not happen and something might be wrong with the file. Or if your application is a target of attacks, or if you want to emulate the original C++ library, you might want to switch to pedantic reading. 

When writing an image, the pedantic flag can be used to skip the validation of your image. For example, if you write an uncompressed file and need maximum speed, it might save a millisecond to disable the checks, if you know that your file is not invalid any ways. I do not recommend this though, as the file might not be readably by any other exr library after that.

In the other compression methods, I think it's only used in the decompression methods, where the buffer is checked after decompression should be over. When bytes are still left even though the decompression is over, and pedantic is true, then an error is returned. If you think this is a simple addition, then you could also do this for B44.

src/compression/b44/mod.rs
johannesvollmer: `memcpy(dest_ptr, src_ptr, count)` can most of the time be replaced with `dest_slice.copy_from_slice(src_slice)`, if you have two slices of the same length. You can create those slices using `let src_slice = &source_slice[start_index .. end_index]`. 

Your code is already doing this. If this is wrong, than using slices might not solve the issue though.

src/compression/b44/mod.rs
johannesvollmer: I think I would replace the indexing with sub slices. By the way, I think you might have mixed up the source and destination?

For example:

```rust
// cpy_u16(&s, 0, &mut tmp_u16_buffer, row0, sample_count);
// instead, I might do:
row0: &mut [u16] = ...;
row1: &mut [u16] = ...;

row0.copy_from_slice(&s[0 .. 4]);
row1.copy_from_slice(&s[4 .. 8]); // or [ 4 .. 4 + sample_count ] ?
row2.copy_from_slice(&s[8 .. 12]);
row3.copy_from_slice(&s[12 .. 16]);
```

However, this would create a borrowing issue, as the different rows probably all come from one array, and you can't simply borrow multiple mutable slices from the same array. There are two possibilities:

A: Borrow the individual rows on the fly every time you want to do anything:
```rust
all_rows[row1_start .. row1_end].copy_from_slice(&s[0 .. 4]);
all_rows[row2_start .. row2_end].copy_from_slice(&s[4 .. 8]); // or [ 4 .. 4 + sample_count ] ?
all_rows[row3_start .. row3_end].copy_from_slice(&s[8 .. 12]);
all_rows[row4_start .. row4_end].copy_from_slice(&s[12 .. 16]);
```

B: My preferred solution: Use slices all the way, avoiding indices wherever possible.
```rust
// two borrow multiple slices out of the same buffer, use one of these approaches:
let (row0, buffer_rest) = buffer.split_mut(row0_end); // assuming all rows immediately follow each other in the buffer
let (row1, buffer_rest) = buffer_rest.split_mut(row1_end); // you can also call this on a subslice of the buffer_rest
let (row2, buffer_rest) = buffer_rest.split_mut(row2_end);
...
```

```rust
// split buffer into multiple slices:
let mut remaining_rows_iter = buffer.chunks_exact_mut(row_size); // split into parts, assuming all rows have the same size 
for (row, index) in remaining_rows.enumerate() { row.copy_from_slice(&s[index * 4 .. index * 4 + count]); }

// you can also extract the iter rows manually without the for loop
let row0 = remaining_rows_iter.next().unwrap();
let row1 = remaining_rows_iter.next().unwrap();
let row2 = remaining_rows_iter.next().unwrap();
```

johannesvollmer:
Some of these changes I suggested should probably be done once we know for sure that the code works correctly, because some changes make it harder to compare it to the C++ code.

src/compression/b44/mod.rs
johannesvollmer: Now that I look at the context, the first version does not make sense, because we never needed to store the temporary value in an array.

```rust
    assert_eq!(b[2], 0xfc);
    let sample = (((b[0] as u32) << 8) | (b[1] as u32)) as u16;
    let sample = if (sample & 0x8000) != 0 { sample & 0x7fff } else { !sample };
    for element in s { *element = sample; }
```

src/compression/b44/mod.rs
```diff
@@ -0,0 +1,924 @@
...
+            for i in channel.tmp_start_index..(channel.tmp_start_index + byte_count) {
+                // TOTALLY WRONG
+                out[out_i + i] = compressed[in_i + i];
+            }
+
+            // Raw copy bytes in u16 temp buffer.
+            for i in (0..byte_count).step_by(2) {
+                let v0 = compressed[in_i + i];
+                let v1 = compressed[in_i + i + 1];
+                tmp_u16_buffer.push(u16::from_be_bytes([v0, v1]))
```
johannesvollmer, line 340: Challenge: Try to copy bytes into the u16 buffer without `for` and without any indexing integer. Here a some hints, start from top to bottom if you are stuck:

- [hint 1](https://doc.rust-lang.org/std/iter/trait.Extend.html)
- [hint 2](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.take)
- [hint 3](https://doc.rust-lang.org/std/primitive.slice.html#method.chunks_exact)

src/compression/b44/mod.rs
Narann: I will take the review one by one. Actually, the whole code hint is now easier to understand than focusing on the loop only:

```rust
// Unpack a 3-byte block `b` into 4 by 4 identical 16-bit pixels in `s` array.
fn unpack3(b: &[u8], s: &mut [u16; 16]) {
    assert_eq!(b[2], 0xfc);

    // Get the pixel value (convert the 2 bytes to u16).
    let mut value = (((b[0] as u32) << 8) | (b[1] as u32)) as u16;

    if (value & 0x8000) != 0 {
        value &= 0x7fff;
    } else {
        value = !value;
    }

    s.fill(value);
}
```

Note the use of `fill()`.

I think this version is more clear: Extract the float value and put it everywhere.

src/compression/b44/mod.rs
Narann: Thanks, the `copy_to_slice()` is the important part.

I think a function make the code simpler to read (I will rename the function, of course). Changing this:

```rust
cpy_u16(&s, 0, &mut tmp_u16_buffer, row0, sample_count);
cpy_u16(&s, 4, &mut tmp_u16_buffer, row1, sample_count);
cpy_u16(&s, 8, &mut tmp_u16_buffer, row2, sample_count);
cpy_u16(&s, 12, &mut tmp_u16_buffer, row3, sample_count);
```
To this:

```rust
tmp_u16_buffer[row0..(row0+sample_count)].copy_from_slice(&s[0..sample_count]);
tmp_u16_buffer[row1..(row1+sample_count)].copy_from_slice(&s[4..(4+sample_count)]);
tmp_u16_buffer[row2..(row2+sample_count)].copy_from_slice(&s[8..(8+sample_count)]);
tmp_u16_buffer[row3..(row3+sample_count)].copy_from_slice(&s[12..(12+sample_count)]);
```

Seems more confusing to me.

`cpy_u16()` is used at some other places.

johannesvollmer: As mentioned in my other suggestions, I would do it very simple:

```rust
row0.copy_from_slice(&s[0 .. 4]);
row1.copy_from_slice(&s[4 .. 8]); // or [ 4 .. 4 + sample_count ] ?
row2.copy_from_slice(&s[8 .. 12]);
row3.copy_from_slice(&s[12 .. 16]);
```

or with a loop? might be more complicated, but maybe you get the idea:
```rust
let sources = s.chunks_exact(4).map(|s| &s[..sample_count]);
let rows = tmp_buffer.chunks_exact(row_length);
for (row, source) in rows.zip(sources) { 
    row.copy_from_slice(source); 
}
```

By trying to work with slices only, never really using indices in the first place. Might be worth a try.

src/compression/b44/mod.rs
Narann: Here, `s` is an `array` (`[u16; 16]`), not a slice. `value` is an `u16`. So `array.fill()` is the proper way.

src/compression/b44/mod.rs
Narann: > And try to have the slices, no indices, throughout the whole algorithm.

Ah, OK, make sense. I will try to see how I can do that.

Narann:
I try to put iterators where I can but find it more confusing sometimes. Do you think this:

```rust
    for (s_elem, t_elem) in s.iter().zip(t.iter_mut()) {
        if (*s_elem & 0x7c00) == 0x7c00 {
            *t_elem = 0x8000;
        } else if (*s_elem & 0x8000) != 0 {
            *t_elem = !*s_elem;
        } else {
            *t_elem = *s_elem | 0x8000;
        }
    }
```

Is nicer than this?

```rust
    for i in 0..16 {
        if (s[i] & 0x7c00) == 0x7c00 {
            t[i] = 0x8000;
        } else if (s[i] & 0x8000) != 0 {
            t[i] = !s[i];
        } else {
            t[i] = s[i] | 0x8000;
        }
    }
```

The visual structure of the second one look clearer to me on what is happening to the array.

src/compression/b44/mod.rs
Narann: I end up with this working:

```rust
tmp_u16_buffer.extend(compressed.chunks_exact(2).map(|c| u16::from_be_bytes([c[0], c[1]])));
```

src/compression/b44/mod.rs
johannesvollmer: Excellent! I wonder why this method is not available for all slices, quite disappointing to have it only for arrays

src/compression/b44/mod.rs
johannesvollmer: Awesome! That's exactly what I had in mind. Of course, whether this is better than the original is still up to debate. It's a rather strange piece of algorithm, but if it works, it aint stupid.

johannesvollmer:
Interesting example. In this case, the second one looks better. 

I'm sad that we can't collect into arrays of fixed size. If this was possible, the following code might be nice:
```rust
let t: [u16; 16] = s.iter()
    .map(|s|
        if s & 0x7c00 == 0x7c00 { 0x8000 }
        else if s & 0x8000 != 0 { !s }
        else { s | 0x8000 }
    )
    .collect();
```
which I think is much clearer. This is how iterator is supposed to work, but `collect` only works for `Vec`, which is a shame. Maybe this will be possible with the new constant integers.

src/compression/b44/mod.rs
johannesvollmer: This code looks like it might be a single call to one of the `lebe` functions. `lebe` also supports the simple conversion from u16 to u8 bytes, even if no endianness conversion is desired. 

Remember that a vec can also be used like any `io::Write` object, which appends the written bytes to the end of the vector

src/compression/b44/mod.rs
Narann: I have multiple troubles with this one.

As B44 is decompressing 4x4 blocks, it have to be initialized once so you can write 4x4 pixels in the buffer. This mean you can't `push()` on it easily. You have to initialize it and write 4x4 blocks on it.

I'm digging with `copy_from_slice()` because I want `memcpy`. I realize there are no proper way to convert a `u16` slice to a `u8` one.

```rust
    let s_slice = &s[s_i..(s_i+n)];  // &[u16]
    let out_slice = &mut out[(row*2)..((row*2)+n*2)];  // &[u8]

    out_slice.copy_from_slice(s_slice); // fail to compile.
```

I suspect the only way to do a real `memcpy` like [the reference implementation](https://github.com/AcademySoftwareFoundation/openexr/blob/master/src/lib/OpenEXR/ImfB44Compressor.cpp#L950) is to have a `unsafe` dedicated to 4 u16 copy to u8 slice. All others will imply a unrolled loop. I can do a loop, we simply miss the `memcpy`. You are the maintainer, you choose. :)

```rust
    let mut out_i = row*2;

    for v in &s[s_i..(s_i+n)] {
        let v_array = v.to_be_bytes();
        out[out_i] = v_array[0];
        out[out_i+1] = v_array[1];
        out_i +=2;
    }
```

johannesvollmer: This is exactly the purpose of the `lebe` library. You can not only use it to convert endianness, but also convert between bytes and u16 slices without changing endianness. 

See here: https://docs.rs/lebe/0.5.1/lebe/io/trait.WriteEndian.html#method.write_as_native_endian

To use this writer method, you can call `std::io::Cursor::new(&mut slice)` to create a writer, which can then overwrite the contents in the `slice` using any IO operation. One kind of IO operation would be to convert endianness, or another would be to simply copy the contents of another slice.
 
Full (untested) example:

```rust
use lebe::prelude::*;
use std::io::Cursor;

fn convert_u16_to_u8(){
    let mut bytes = vec![0_u8; 1000]; // allocate 1000 bytes output
    let u16_values = vec![ 2343; 32 ]; // our 32 numbers 
    
    let mut bytes_writer = Cursor::new(&mut bytes[2 .. ]); // overwrite the contents of the byte vector, starting at the third byte
    bytes_writer.write_as_native_endian(u16_values).unwrap(); // overwrite 32*sizeof(u16) bytes, internally using a memcpy, without converting endianness, the first three bytes are not touched
}

```

Narann: Thanks, I looked a little into the `lebe` libs, I missed the `Cursor`, but it still worry me. Does not `Cursor::new()` allocate on the heap?

I cant test on playground rust, but I suspect this `convert_u16_to_u8()` will not be compiled to `memcpy` instruction only because of the heap allocation. Allocating on heap kinda lose the benefit of `memcpy` IMHO.

I also suspect `lebe` to focus on write/read on/from IO stream of byte. Here we are manipulating in-memory arrays/slice.

johannesvollmer: > Does not Cursor::new() allocate on the heap?

No, I don't think so. Actually, the slice on its own already implements `Read`. Technically we don't need the cursor to use a vector like a file, but it delivers intent. It does not allocate if you pass it a slice. It only allocates if you pass it a vec and the vec needs to allocate.

> I also suspect lebe to focus on write/read on/from IO stream of byte. Here we are manipulating in-memory arrays/slice.

The purpose of lebe is twofold: A) to make converting between bytes and slices possible without the use of unsafe, B) be endian-aware while doing that.

The Rust `Read` and `Write` trait are not restricted to files. They are often used for in-memory manipulation by design. They will always compile to a memcpy in that situation. 

Doing conversion from u32/f32/f16 to bytes is the main goal of lebe, and it does not matter whether your data is in memory or being written to a physical file. :)

src/compression/b44/mod.rs
Narann: > which appends the written bytes to the end of the vector

This is my current limitation, as B44 manipulate blocks of 4x4. When I decode them, I need the final image buffer to write each of the 4 rows.

johannesvollmer: I don't know how the data is structured. I assume all four rows are stored one after another in a large buffer? Maybe it would make sense to split the buffer into the four rows, using either `buffer.chunks_mut(row_length)` or calling `buffer.split_at_mut(nect_row_end)`. This way you can obtain four slices which can be used to write the four rows simultaneously.

src/compression/b44/mod.rs
Narann: After more investigation, I don't think I can skip the [use of a temp buffer](https://github.com/AcademySoftwareFoundation/openexr/blob/master/src/lib/OpenEXR/ImfB44Compressor.cpp#L866). It seems the b44 uncompress implementation first uncompress block in the order they are compressed. My efforts to remove the use of a temp buffer to directly write inside the output buffer would means a big move away from the reference implementation. I think it's too soon.

I suggest to keep the current implementation (use a u16 buffer, then convert to scanline u8 buffer), and continue to clean the code and don't touch to the implementation.

johannesvollmer: Sounds good. Optimization should be the last step, if there is no simple solution. :)

src/compression/b44/mod.rs
Narann:
```rust
s[0..4].copy_from_slice(&tmp[row0..(row0+4)]);
s[4..8].copy_from_slice(&tmp[row1..(row1+4)]);
s[8..12].copy_from_slice(&tmp[row2..(row2+4)]);
s[12..16].copy_from_slice(&tmp[row3..(row3+4)]);
```

Narann:
I've kinda cleaned up the code. Use iterator when it make sense. I see some possible improvement (compute the uncompressed temp buffer size once, at the beggening and don't use `resize()`).

For now, There is one TODO (handle "compression" of F32 and U32 buffers) and I want to focus on tests. As you can see, my tests are bad (write B44 image, read it, write it uncompressed, then B44 compressed). I need to write more of them. Any place in the code where you consider are "good tests I should look at and reproduce?".

We are not that far. :)

Thanks in advance.

src/compression/b44/mod.rs
johannesvollmer: How about `tmp_bytes[..].read_from_big_endian_into(&mut u16_buffer[..])`?

johannesvollmer:
Nice! 

I recommend looking at the PIZ tests, the other ones don't have very complicated tests. And if you can find some "ground truth" compressed/uncompressed data in the original C++ repository, that would be fantastic. Also, you can generate random values to compress and decompress.

Usually, we test like this: read valid B44 file generated by another program, compress the whole file to an in-memory byte vector, then uncompress these bytes again, and check whether it is similar. Most of the existing B44 images will not be supported because of other missing features, so you might have to create some new B44 files. Please put these images into the `exrs/tests/images/valid/custom` and then use a relative path. Please don't write to a temporary file, as you can just use a plain vector.

johannesvollmer:
> TODO (handle "compression" of F32 and U32 buffers) 

I thought these types are not compressed? Did you not add the code to copy them?

Narann:
Yes, I need to add the copy code (the reason why I used double quotes for "compression").

So:

* [ ] Generate random values, compress, decompress, test for almost equal.
* [ ] Generate 2 images from another program (both compressed and not), store them in `exrs/tests/images/valid/custom`. Open and compress the original image and check absolute equality.

I need to do this for both B44 and B44A.

Seems like a good plan.

johannesvollmer:
I only merge the `dev` branch into yours because I'm sure these files will not affect you directly :)

johannesvollmer:
Does merging the dev branch result in any inconveniences for you? I'll not push the branch then, you can just pull yourself when you are ready.

Narann:
No problem, I simply rebase on `dev` everytime you make any change so I'm sure there will be no surprise during final merging.

Narann:
The implementation highly relies on using a preallocated block of memory and using typed pointers (u8, u16) to read/write on them.

I've removed the use of u16 slices in `decompress()` to only relies on u8 slices. This simplifies the f32/u32 handling, but make the rest on the code a little bit ugly, and, I suspect, inefficient.

Here is the discussion about having converting a u16 slice to a u8 one: https://users.rust-lang.org/t/u16-slice-to-u8-iterator/59078

The final assembly code is not as simple as it should be IMHO.

I will continue to totally remove the use of u16 slices in the buffer, but I think we will need a better `memcpy` than:

```rust
    for src_v in &src[src_i..(src_i+n)] {
        for b in &src_v.to_ne_bytes() {
            dst[i] = *b;
            i += 1;
        }
    }
```

johannesvollmer:
The code you mentioned:

```rust
    for src_v in &src[src_i..(src_i+n)] {
        for b in &src_v.to_ne_bytes() {
            dst[i] = *b;
            i += 1;
        }
    }
```

Can be perfectly replaced with 
```rust
dst[i .. i+2*n].write_as_native_endian(&src[src_i..src_i+n])
```

Which will use a true `memcpy`

Narann:
Thanks, TBH anytime I tried to use `lebe`, it didn't work. I tried again:
```rust
fn cpy_u8(src: &[u16], src_i: usize, dst: &mut [u8], dst_i: usize, n: usize) {

    // let mut i = dst_i;
    //
    // for src_v in &src[src_i..(src_i+n)] {
    //     for b in &src_v.to_ne_bytes() {
    //         dst[i] = *b;
    //         i += 1;
    //     }
    // }
    // debug_assert_eq!((i - dst_i) / 2, n);

    use lebe::io::WriteEndian;
    dst[dst_i..dst_i+2*n].write_as_native_endian(&src[src_i..src_i+n]).unwrap()
}
```

Error message:

``` text
error[E0599]: no method named `write_as_native_endian` found for slice `[u8]` in the current scope
   --> src/compression/b44/mod.rs:257:27
    |
257 |     dst[dst_i..dst_i+2*n].write_as_native_endian(&src[src_i..src_i+n]).unwrap()
    |                           ^^^^^^^^^^^^^^^^^^^^^^ method not found in `[u8]`
```

johannesvollmer:
Okay, good to know, there's no real problem here. The only problem is that I takez this is my phone and didn't actually test this (sorry).

Here's a real example:
https://github.com/johannesvollmer/exrs/blob/c4a591f4b21580f5d4478455e7a28416ef770e80/src/compression/piz/mod.rs#L196

Maybe you'll have to add `use lebe::...` or have I didn't recall the correct method name. Try the one from the example :)

johannesvollmer:
Oh, I see the `use lebe` statement. Maybe we need another `use std::io::Write` import? I can try to find the exact problem once I have time for my computer again. Might be today in the evening

Narann:
I still need to test, but [this](https://godbolt.org/z/cjYE564E8) seems to be the best implementation:

```rust
pub fn u16_to_u8_cpy(src: &[u16], dst: &mut [u8]) {
    src.iter()
        .zip(dst.chunks_exact_mut(2))
        .for_each(|(a, b)| b.copy_from_slice(&a.to_ne_bytes()));
}
```
It generate the proper ASM:
```asm
example::u16_to_u8_cpy:
        push    rax
        shr     rcx
        cmp     rcx, rsi
        cmovae  rcx, rsi
        test    rcx, rcx
        je      .LBB0_2
        mov     rax, rdi
        add     rcx, rcx
        mov     rdi, rdx
        mov     rsi, rax
        mov     rdx, rcx
        call    qword ptr [rip + memcpy@GOTPCREL]
.LBB0_2:
        pop     rax
        ret
```

johannesvollmer:
> proper ASM

That would absolutely work. I'll be glad to merge this when works and is safe, as we can always optimize later. :)

Narann:
Big code cleanup, rustified few loops, use byte buffers everywhere instead of u16 buffers, add a test for mixed type channels, still some print to remove and more tests to add and it should be ready for merge.

Narann:
My last task is now to fix the last failing test: `compare_compression_content` with `f16/b44.exr` image. Need investigate what the problem is.

Narann:
I've opened a ticket, but I'm skeptical about the efficiency of recursive validation using the `ValidateValueResult` trait. Because there is only one assert for the final result of `validate_image_result()`, I need to put `println!()` in every `validate_image_result()` implementation to identify where things fail.

Maybe we should use assert in each check of `validate_image_result()` instead of create a whole condition.

johannesvollmer:
I see the problem. Yes, we can use `assert` instead of `return false`, which allows us to have more information about what comparison failed exactly. 

Alternatively, we could also return an error object, which could additionally record which header and which channel and which sample index contains the failed comparison. I assume a plain assert will not help in this regard, it will only tell you "float a does not equal float b", without knowing where we are. 

Third option: This could be the perfect opportunity to try a Debugger, it's made exactly for that purpose. I think VS Studio has a a binary Debugger

Narann:
Error object is a good idea.

I have find *where* the problem is:

```rust
    impl ValidateValueResult for f16 {
        fn validate_value_result(&self, other: &Self, max_difference: Option<f32>, nan_to_zero: bool) -> bool {
            print!("self other to_f32 {} {}\n", self, other);
            print!("self other to_f32 {} {}\n", self.to_f32(), other.to_f32());
            self.to_f32().validate_value_result(&other.to_f32(), max_difference, nan_to_zero)
        }
    }
```

Log is:

```text
self other to_f32 5.4257813 5.0703125
self other to_f32 5.4257813 5.0703125
```

So I suspect there is a problem with high value. Pixel coordinate would help

Narann:
I did a difference, in GIMP between `tests/images/valid/custom/compression_methods/f16/uncompressed.exr` and `tests/images/valid/custom/compression_methods/f16/b44.exr`:

image

I will try with another software, but I suspect my code is correct, but B44 compression have big impact on pixels with high values.

Narann:
I tried with Nuke. Here is the value at the center of the sphere:

image

When I do a difference I have big offsets around the circle:

image

image

I tried to write the uncompressed image to a B44 compression with Nuke and compare it to uncompressed and have the same problem.

So it doesn't seems to be related to my implementation of B44, but B44 seems to loss precision in very high values.

I will do a final cleanup for the rest of the code and I think it should be ready to merge.

johannesvollmer:
Just adjust the threshold in the test in `across_compression.rs` on line 34, for example: 
`let is_similar = uncompressed.validate_image_result(&decompressed, 0.1);` instead of `0.0001`

johannesvollmer:
`5.4257813` to `5.0703125` is quite a big difference, I don't know whether this much of a difference is to be expected from B44. If there would be a problem in the code, it would probably be a bigger difference though, I assume. 


Can you create an uncompressed and B44 file in Nuke/Gimp and compare the difference between these two images? The difference between officially compressed images could be the same difference as in the Rust library.

johannesvollmer:
When you are satisfied with the code, hit the `Ready for review` button below, to make this a real pull request.

Narann:
> Can you create an uncompressed and B44 file in Nuke/Gimp and compare the difference between these two images? The difference between officially compressed images could be the same difference as in the Rust library.

I already did it:

> I tried to write the uncompressed image to a B44 compression with Nuke and compare it to uncompressed and have the same problem.

I took `tests/images/valid/custom/compression_methods/f16/uncompressed.exr` and write it to `tests/images/valid/custom/compression_methods/f16/b44_nuke.exr`, then read it and do a difference between them:

image

`Difference2` and `Difference1` have the exacte same output. `Difference3` compare `B44` generated from Nuke to `B44` generated by me with this code:

```rust
    fn b44_big() {
        let mut image = read_first_flat_layer_from_file(
            "/mnt/storage/code/repo/exrs/tests/images/valid/custom/compression_methods/f16/b44.exr").unwrap();

        image.layer_data.encoding.compression = crate::compression::Compression::Uncompressed;
        image
            .write()
            .to_file("/home/user/Desktop/comp_f16_b44_uncomp.exr")
            .unwrap();

        let mut image = read_first_flat_layer_from_file(
            "/mnt/storage/code/repo/exrs/tests/images/valid/custom/compression_methods/f16/b44.exr").unwrap();
        image.layer_data.encoding.compression = crate::compression::Compression::B44;
        image
            .write()
            .to_file("/home/user/Desktop/comp_f16_b44_b44.exr")
            .unwrap();
    }
```

There is no difference. So I guess my output is correct.

johannesvollmer:
If the difference you observe is `1000.0`, I assume there are very large values in the image, right? 

Currently the tests uses a simple absolute difference for checking the equality. As this does not seem to work, we should consider using a relative difference. See the second answer on [this stack overflow thread](https://stackoverflow.com/questions/4915462/how-should-i-do-floating-point-comparison). Will this fix the problem?

johannesvollmer:
I implemented a relative comparison for floats instead of absolute comparison. I added tests so I'm pretty sure it will work. Use a threshold of `0.7` or similar.

src/compression/b44/mod.rs
```diff
@@ -0,0 +1,863 @@
...
+        let rectangle = IntegerBounds {
+            position: Vec2(-30, 100),
+            size: Vec2(322, 731),
+        };
+
+        test_roundtrip_noise_with(channels, rectangle);
+    }
+
+    #[test]
+    fn roundtrip_any_sample_type_toto() {
```
johannesvollmer, line 800: Please don't use absolute paths in the tests. I suggest you either remove this test, or write to an in-memory byte vec instead of a file.

Narann: This one will disappear.

Narann:
I consider the code ready for review.

`compare_compression_contents` test still fail, but I think it's a different problem.

```
f16/b44.exr: thread 'compare_compression_contents' panicked at 'b44 compression does not match uncompressed', tests/across_compression.rs:35:21
```

johannesvollmer:
Excellent. I'll add a basic implementation for the error object, to simplify finding the exact problem. You can always add more information to that basic implementation where you need it.

johannesvollmer:
So... I did that thing

Narann:
> So, is the problem tracked down already?

Yes, the `compare_compression_content` is not working (meaning, it can't work anyway) on high very values. This message came from the Github build:

```
--- compare_compression_contents_b44a_f16 stdout ----
f16/b44a.exr: checking b44a compression for equality to uncompressed data
thread 'compare_compression_contents_b44a_f16' panicked at 'called `Result::unwrap()` on an `Err` value: "| image > layer data (layer `None`) > channel `B` > sample_data > f16 samples element [42291] of 2073600: expected ~16.421875, found 20.28125 (adaptive tolerance 2.2021875)"', /home/runner/work/exrs/exrs/src/image/mod.rs:887:87
```

By the way, now we have isolated tests, we can also provides per-test compare information (lossless, lossy).

johannesvollmer:
Ah, I see. Merging would be possible, but I require that all test run. 

Solving the test: We could work with an per-call threshold, yes. The lossless methods are automatically detected and are always tested perfect equality, so no worries about them.

Regarding B44, there is another thing we could do:
Using nuke or another program, import the (already slightly different) B44 compressed file, and re-export the already changed values without compression as `expected_decompressed_b44_result.exr`. Then we can test for perfect equality between the real B44 and "uncompressed B44". With this, we could be 100% sure, and avoid approximate comparison entirely. (Just be sure to set all conpression_method fields to compression::uncompressed before comparing)

Also, we can test writing the normal uncompressed file as B44 file, and check whether the result equals `expected_decompressed_b44.exr`.

I'll not be able to use my computer until Monday, so you could implement that test yourself.

johannesvollmer:
Modified the tests, saved the b44 files in the directory as uncompressed files. B44 f32 is compared to the uncompressed, as f32 are not compressed at all (still funny lol).

johannesvollmer:
The error occurs when a B44 file is read, then written with the same compression, and then read again. The two read images do not match. This is actually not okay, for the simple reason that the original b44 file already had compression artefacts, and recompressing the b44 file should probably produce the same b44 file again, with the same compression artefacts, right?

johannesvollmer:
image

Looks like either the reader or writer got confused here and put quite random pixels in the last three columns of the image (maybe this issue only appears at this specific resolution).

image

Narann:
Thanks, for the investigation, I can reproduce the bug, I will try to find the proper behavior.

(I suggest to never use random values for image dimension test, but detect the case causing troubles and do a dedicated test for it.)

Narann:
So, I technically fixed the problem, it was an error in `j` calculation in image border:

```rust
                if x + 3 >= x_sample_count {
                    let n = x_sample_count - x;

                    for i in 0..BLOCK_SAMPLE_COUNT {
                        // let j = min(i, n - 1);
                        let j = min(i, n - 1) * 2;  // we manipulate bytes, not u16

                        // TODO: Make [u8; 2] to u16 fast.
                        s[i + 0] = u16::from_ne_bytes([tmp[(row0 + j)], tmp[(row0 + j + 1)]]);
                        s[i + 4] = u16::from_ne_bytes([tmp[(row1 + j)], tmp[(row1 + j + 1)]]);
                        s[i + 8] = u16::from_ne_bytes([tmp[(row2 + j)], tmp[(row2 + j + 1)]]);
                        s[i + 12] = u16::from_ne_bytes([tmp[(row3 + j)], tmp[(row3 + j + 1)]]);
                    }
                } else {
```
But I can't reproduce the problem. We still need some investigation to provide a reproductible

johannesvollmer:
What do you mean you can't reproduce it? Do you mean that you tried adding tests, but the tests never failed, even with the old code?

johannesvollmer:
This is the problematic image meta data, simplified:

```
Header {
    channels: ChannelList {
        list: [
            ChannelDescription {
                name: exr::Text("B"),
                sample_type: F16,
                quantize_linearly: false,
            },
            ChannelDescription {
                name: exr::Text("G"),
                sample_type: F16,
                quantize_linearly: false,
            },
            ChannelDescription {
                name: exr::Text("R"),
                sample_type: F16,
                quantize_linearly: false,
            },
            ChannelDescription {
                name: exr::Text("left.B"),
                sample_type: F16,
                quantize_linearly: false,
            },
            ChannelDescription {
                name: exr::Text("left.G"),
                sample_type: F16,
                quantize_linearly: false,
            },
            ChannelDescription {
                name: exr::Text("left.R"),
                sample_type: F16,
                quantize_linearly: false,
            },
            ChannelDescription {
                name: exr::Text("right.B"),
                sample_type: F16,
                quantize_linearly: false,
            },
            ChannelDescription {
                name: exr::Text("right.G"),
                sample_type: F16,
                quantize_linearly: false,
            },
            ChannelDescription {
                name: exr::Text("right.R"),
                sample_type: F16,
                quantize_linearly: false,
            },
        ],
        bytes_per_pixel: 18,
        uniform_sample_type: Some(F16),
    },
    compression: B44A,
    blocks: ScanLines,
    line_order: Increasing,
    layer_size: Vec2(775, 678),
    deep: false,
    deep_data_version: None,
    chunk_count: 22,
    max_samples_per_pixel: None,
    shared_attributes: ImageAttributes {
        display_window: IntegerBounds {
            position: Vec2(0, 0),
            size: Vec2(775, 678),
        },
        pixel_aspect: 1.0,
        chromaticities: None,
        time_code: None,
        other: {},
    },
    own_attributes: LayerAttributes (default values omitted) {
        name: None,
        horizontal_density: Some(100.0),
        comments: Some(exr::Text("Plane blade adjustment mechanism")),
        multi_view_names: Some(
            [
                exr::Text("center"),
                exr::Text("left"),
                exr::Text("right"),
            ],
        ),
    },
}

```

Narann:
> What do you mean you can't reproduce it?

Sorry, I mean I can't reproduce a test that generate the problem.

> Do you mean that you tried adding tests, but the tests never failed, even with the old code?

Yes, for a unknown reason, test with random color values doesn't trigger the problem.

> With your change, all tests pass on my machine now.

That's a good news!

> Interesting, I too can't make a test fail with these settings, strange

Thanks, I'm not dumb, haha.

As the image is inside the repository we could simply use it as a real test, maybe?

johannesvollmer:
> As the image is inside the repository we could simply use it as a real test, maybe?

Yeah, `cargo test` already uses all images in the repository, including the exact B44 image causing the problem, for a simple round trip test. It would be nice to have it as an explicit test in the b44 module though, such that the test fails if that image is accidentally removed from the repository.

Narann:
> Still it would be nice to have it as an explicit test in the b44 module though

Just to be sure: You ask me to do a test in b44 with explicit path to this image file right?

johannesvollmer:
Yes. 

First, run `cargo test`. This goes through all images in the repo and runs some tests with it. Go to the test which test fails (probably `roundtrip.rs:test_roundrip_all_files_full`). Copy the test, but only the code that runs for a single image. Create a new test in the b44 module that loads the specific file that failed (probably `Adjuster.exr`). 

With the bug in the code, run your new test once to make sure that it would reveal the bug. Then, once you know the test works, implement the bug fix that you previously posted in the comments and test again to make sure that it now works :)

I would be happy if you implement that test yourself, because I think it would nice if you are the one to finish this whole story :)

Narann:
And finally. I did the test without and with the fix (it respectively break and pass). You can try to remove the `* 2` in line: https://github.com/johannesvollmer/exrs/pull/112/commits/552883de53095ad9633b0efe297f09a71ee2d194#diff-a80e527afb35f1f53f9cbf7e8c42271fbac87d819a2e167c0d60f417363d5b6eR604

I also improved test to compare to original data when it make sense. :D

johannesvollmer:
I hate to do that, but I have to tell you that the github test suite failed. It says "file not found" for the new test.

src/compression/b44/mod.rs
```diff
@@ -0,0 +1,964 @@
...
+            test_roundtrip_noise_with(channels, rectangle);
+
+        assert_eq!(pixel_bytes.len(), 60);
+        assert_eq!(compressed.len(), 62);
+        assert_eq!(decompressed.len(), 60);
+    }
+
+    #[test]
+    fn border_on_multiview() {
+        // This test is hard to reproduce, so we use the direct image.
```
johannesvollmer, line 938: Instead of the absolute path ` "/mnt/storage/code/repo/exrs/tests/images/valid/openexr/MultiView/Adjuster.exr"`, try using the relative path ` "tests/images/valid/openexr/MultiView/Adjuster.exr"`

johannesvollmer (commented):
Whoopsie

src/compression/b44/mod.rs
Narann: Facepalm… Too much enthusiasm from me.

Narann:
:D

src/compression/b44/mod.rs
johannesvollmer: No worries :) I think we can merge now

johannesvollmer:
:D

Alright the tests have passed, let's meeeeerge

johannesvollmer:
Woohoo 🙌 Fantastic! Thanks for your patience, Dorian :)

Narann:
Great! Thanks for your feedbacks! :D

PR #113, opened 2021-04-16
change assert to debug_assert

Narann:
assert!() macros are always checked in both debug and release builds, and cannot be disabled, so we change them for debug_assert!() macros.

https://doc.rust-lang.org/std/macro.assert.html

I don't see any reason to keep those asserts in release builds (but maybe there is a good one so feel free to close the PR).

### DIFF
--- a/src/compression/mod.rs
+++ b/src/compression/mod.rs
@@ -166,8 +166,8 @@ impl Compression {
     pub fn compress_image_section(self, header: &Header, mut uncompressed: ByteVec, pixel_section: IntegerBounds) -> Result<ByteVec> {
         let max_tile_size = header.max_block_pixel_size();
 
-        assert!(pixel_section.validate(Some(max_tile_size)).is_ok(), "decompress tile coordinate bug");
-        if header.deep { assert!(self.supports_deep_data()) }
+        debug_assert!(pixel_section.validate(Some(max_tile_size)).is_ok(), "decompress tile coordinate bug");
+        if header.deep { debug_assert!(self.supports_deep_data()) }
 
         // convert data if compression method expects native format
         // see https://github.com/AcademySoftwareFoundation/openexr/blob/3bd93f85bcb74c77255f28cdbb913fdbfbb39dfe/OpenEXR/IlmImf/ImfTiledOutputFile.cpp#L750-L842
@@ -204,8 +204,8 @@ impl Compression {
     pub fn decompress_image_section(self, header: &Header, compressed: ByteVec, pixel_section: IntegerBounds, pedantic: bool) -> Result<ByteVec> {
         let max_tile_size = header.max_block_pixel_size();
 
-        assert!(pixel_section.validate(Some(max_tile_size)).is_ok(), "decompress tile coordinate bug");
-        if header.deep { assert!(self.supports_deep_data()) }
+        debug_assert!(pixel_section.validate(Some(max_tile_size)).is_ok(), "decompress tile coordinate bug");
+        if header.deep { debug_assert!(self.supports_deep_data()) }
 
         let expected_byte_size = pixel_section.size.area() * header.channels.bytes_per_pixel; // FIXME this needs to account for subsampling anywhere
 
--- a/benches/write.rs
+++ b/benches/write.rs
@@ -47,7 +47,7 @@ fn write_parallel_zip16_to_buffered(bench: &mut Bencher) {
 fn write_uncompressed_to_buffered(bench: &mut Bencher) {
     let path = "tests/images/valid/custom/crowskull/crow_uncompressed.exr";
     let image = read_all_flat_layers_from_file(path).unwrap();
-    assert!(image.layer_data.iter().all(|layer| layer.encoding.compression == Compression::Uncompressed));
+    debug_assert!(image.layer_data.iter().all(|layer| layer.encoding.compression == Compression::Uncompressed));
 
     bench.iter(||{
         let mut result = Vec::new();


johannesvollmer:
In Tests and in benchmarks, we always use assert, as they are not run in release mode anyways.

In the main code, we use assert per default, to catch errors as early as possible. For validating the data a user wants to write, we use assertions so that there is no possibility of writing an invalid file, even in release mode. Only very expensive checks and double-checks (to catch bugs in exrs) are done with `debug_assert`. 

I'll have a look at your changes so that we can find out which is the right one for each situation :)

Narann:
OK, make sense.

The 


> In Tests and in benchmarks, we always use assert, as they are not run in release mode anyways.

I just realize one of my modification is in `benches`, I agree. Those are not build in `--release` and can be left.

> In the main code, we use assert per default, to catch errors as early as possible. For validating the data a user wants to write, we use assertions so that there is no possibility of writing an invalid file, even in release mode.

I politely disagree with using assert in this context. Those errors should be part of the error management of the library, because it makes library third party tools unable to catch the error and properly inform the user. I *think* using of `Result` is a better way to handle this.

Thanks for the clarification, I will remove the `benches` modifications and see how we could avoid the use of assert in other contexts.

johannesvollmer:
> using of Result is a better way to handle this

I agree. This is how it is handled almost everywhere. 

I've had a look at the assert statements in this pull request. If I remember correctly, these are double-checks. They would have resulted in an error if the user did anything wrong. These are just here in case the library has a bug and somehow no error was returned earlier. They are not only debug assertions to avoid writing an invalid file, even if the library has a bug. Also, I think these are very cheap checks, so the performance is probably not a problem here. Correct me if I'm wrong.

However, I think there is actually a place in this library, where assert statements could be replaced with errors. Generally, when writing a file, if anything goes wrong, an error is returned. However, if the user passes obviously wrong data (two vectors lengths are not the same) there will only be an assertion, and not an error (as assertions can usually provide useful debugging information to the user, such as a backtrace). These places should maybe use error objects instead of an assert statement.

johannesvollmer:
In general, I have included many assert statements, not only to catch bugs in the library, but also to convey intent and document the situation. In that sense, it's like adding a comment to your code that says `at this point we know that the pixels are not deep data`. If you want to turn these into an error, there will be multiple places that throw the same error, even though only one of them should ever be thrown if there is not bug in this library. For this reason, there are a ton of assertion and debug assertions scattered everywhere, more than technically would be required. 

So, in general, if you think that an assertions might actually be triggered, maybe because an error should have been thrown but wasn't, then we should make sure that error is thrown somewhere earlier. Or if you find out that the assertion has a significant performance impact, we can change it to a debug assertion. 

Now, the specific assertions in this pull request should never be triggered, as an error should be returned earlier in the process. However, if someone uses the `compress` or `decompress` function directly, and calls it with invalid arguments, they will of course not receive an error but only the simple assertion. Because these functions can be used directly, I prefer real assertions over debug assertions. If you think someone is actually interested in using these raw functions directly, we can consider returning an error instead of having the assertion. In the past, I assumed nobody will use these internal functions, so I did not spend the extra time to use a real error object, and only used a simple assertion. In the end, there is no real disadvantage with using error objects, only that the time we spend could be spent somewhere else, and the error backtrace will be lost. So if you really want to, I see no problem with converting these assertions to an error object.


On a side note: I don't want to demotivate you. And I feel bad for writing such a long paragraph, especially for such a seemingly simple change. I really appreciate that you spend your time contributing to the library, thank you! Might I suggest that we use some kind of real-time messaging app? This way we could first discuss and then code, and not code and then discuss. How about Slack or Discord?

Narann:
Hello and thanks for taking the time to answer.

> They would have resulted in an error if the user did anything wrong.

Yes, but an assert `panic!`, and an error is manageable by the caller, that's the biggest difference. I think a library should never `panic!` if it can be avoided. As a dev using a library, it would make me sad my program to crash (or doing non-sense to avoid to be affected by my library `panic!` call) where I could have managed the error and return a nice message/window to my user.

If the library (`exr` realize it will write an invalid file, it should return a manageable error).

> Also, I think these are very cheap checks, so the performance is probably not a problem here. Correct me if I'm wrong.

I don't think there are any performance problem involved (except a traceback report code/string send with release binary for no reason, which is sad).

> These places should maybe use error objects instead of an assert statement.

I agree. Once again, the problem is mostly an API problem. A library, have users which are not the end user (a dev that will write a nice UI for example). Asserts don't help them.

> In general, I have included many assert statements, not only to catch bugs in the library, but also to convey intent and document the situation. In that sense, it's like adding a comment to your code that says at this point we know that the pixels are not deep data. If you want to turn these into an error, there will be multiple places that throw the same error, even though only one of them should ever be thrown if there is not bug in this library. For this reason, there are a ton of assertion and debug assertions scattered everywhere, more than technically would be required.

So you have a friend here. ;)

I am an big user of asserts and I share the same vision: Asserts are a way to comment the logic of the code ("Here, this *must* be true. If you don't think so, you don't understand this code, or I missed something.").

But I keep my point: If the library user can trigger an assert (and so, lost control of it's program because of the library he call, in this case `exr`), this means you have to tell him (in the documentation) its his own responsability to provide good things. In this situation, this mean you tell the devellopper to check the user inputs are consistent. I suspect the dev prefer to handle and error than think about OpenEXR file format details.

I don't know if I'm clean. I think it's more usefull to document and error the library user could catch, than asking the dev to rely on implementation details (dev maybe use `exr` because he don't know OpenEXR details).

It's mostly a API/responsability discussion than a performance one.


> Now, the specific assertions in this pull request should never be triggered, as an error should be returned earlier in the process.

In my opinion, that's a reason for not letting this in release, and so use `debug_assert`.

> However, if someone uses the compress or decompress function directly, and calls it with invalid arguments, they will of course not receive an error but only the simple assertion.

I think that the point here: If you don't want a user to use a low level function, hide it. If you want to let the user use a low level function without the bound check: Tell him (doc) and leave it alone (no assert, and yes, possible corrupt file, that's the water he choosed to walk on), but don't make the legitimate user that spend time on error management in a uncatchable situation.

> In the end, there is no real disadvantage with using error objects, only that the time we spend could be spent somewhere else,

That's a valid point for explaining the current situation, but it leaves the error management question valid. For now, if I don't want to panic!, I have to check my inputs in a way OpenEXR support it (keep in mind I have maybe no idea of the internals on OpenEXR), and cross my finger for not triggering the `panic!`. I still think a library should document the errors it will return and let the caller handle those errors.

> and the error backtrace will be lost.

That's a problem for *us* (the library maintainers), but it's not the probleme of your library user. The problem of the library user is to keep control of the program and not having it crash in front of its user. `panic!` should be use when you can't go back, not when you can't deliver.

https://doc.rust-lang.org/std/macro.panic.html

It's like you would force your user to have an unwrap inside your code, an unwrap he didn't choose.

Also, the backtrace management logic and strings will be ship inside the binary (useless strings).

> So if you really want to, I see no problem with converting these assertions to an error object.

(Just a note: I'm in favor of letting the `benches/write.rs` with simple `assert!`, as benches are not shipped in `--release` mode. (are they?)).

If you don't want to have a `Result` return, you actually can let the code fly if it's documented directly: This function assume that "blah blah and will generate corrupted file if it's not." and have a `debug_assert!`.

But in those situations the returned arguments are `Result<ByteVec>` so it make sense to return an error.

> On a side note: I don't want to demotivate you. And I feel bad for writing such a long paragraph, especially for such a seemingly simple change.

Don't worry, I am used to those kind of disscussions. You are the maintainer, you do what you think is right. Saying "no" is hard, but valuable (mostly when someone came with 10+ dependencies to support a corner case, but, hey, that's not what we discuss about). If I can't convince you, after this, I will not chime or what. I just try to propose what I think is the best thing to do, like anyone, but it's your baby, and you can totally say: I see your point, but I don't want to push this project in this direction. That not a problem.

> I really appreciate that you spend your time contributing to the library, thank you!

I like rust and work in the CGI so I actually enjoy writting code for `exr`, so thanks *you* for the orginal effort! :)

Don't think too much about me as a person but the project, the best think you could do for me (and maybe others) is to simply spend some time thinking about what I say, letting some time and you will know what to do.

> Might I suggest that we use some kind of real-time messaging app? This way we could first discuss and then code, and not code and then discuss. How about Slack or Discord?

(Note: I did this PR as a start of a discussion, I tend to do like this for simple modifications, no time lost, no worry)

My current situation make me hard to be available. I really wanted to be much more active, but, hey, life. So I prefer those static discussions, if you don't mind. It make my time easier to manage, hope it's OK for you. You can contact me at `fevrier.dorian[at]yahoo.fr`.

johannesvollmer:
Convincing points. Especially where `Result` is already returned, it would be no effort to make it an error object. The specific code you submitted in this PR will be one of these cases. I guess I just never considered that anyone used these functions directly, which could alternatively make `pub(crate)` a good solution. 

> no assert, and yes, possible corrupt file, that's the water he choosed to walk on.

If we remove an assert, a different assert later on might still trigger, maybe a simple array lookup, which will then again not result in an valid or invalid file. I assume it would be an astronomical effort to guarantee never triggering `panic!` in the whole library, even for invalid dev input. It's easier to make the guarantee `will panic when invalid` instead of `will never panic even though invalid`. Saying `might panic or might not panic when invalid` seems to be of no value to me. Therefore, if forced to make a decision, I would choose `will panic when invalid`.

> panic! should be use when you can't go back, not when you can't deliver

As a programmer, I agree, because this is what the designers of Rust planned as the purpose for a panic. As an app user, I see it this way: When I use Affinity Photo, I'd much rather have it crash immediately than write an invalid file, because I might not even notice the file being invalid until it's too late. This might not apply to all apps that will use this library, but I assume I would prefer this in most apps.

> I just try to propose what I think is the best thing to do

I appreciate your input, thanks again :)

### Next Steps

In general, for the whole crate, these are my favourite two options:
- return result instead of asserting 
- make the functions `pub(crate)` 

I think the second option would be beneficial in the long run, reducing API surface. However, I would not do this without making many other functions `pub(crate)` that could be rather internal. So I would choose the first option, probably accepting this PR without any further changes

src/compression/mod.rs
```diff
@@ -166,8 +166,8 @@ impl Compression {
     pub fn compress_image_section(self, header: &Header, mut uncompressed: ByteVec, pixel_section: IntegerBounds) -> Result<ByteVec> {
         let max_tile_size = header.max_block_pixel_size();
 
-        assert!(pixel_section.validate(Some(max_tile_size)).is_ok(), "decompress tile coordinate bug");
-        if header.deep { assert!(self.supports_deep_data()) }
+        debug_assert!(pixel_section.validate(Some(max_tile_size)).is_ok(), "decompress tile coordinate bug");
```
johannesvollmer, line 169: Ok to not have in release mode, because low level api and because not exr specific

src/compression/mod.rs
```diff
@@ -166,8 +166,8 @@ impl Compression {
...
+        if header.deep { debug_assert!(self.supports_deep_data()) }
```
johannesvollmer, line 170: Ok to not have in release mode, because low level api and will change in the future

johannesvollmer (changes requested):
Good to go (except for benches of course). Waiting for your reply before merging though. 

Thanks for the number of good points you contributed, they will improve the library :)

Narann:
> When I use Affinity Photo, I'd much rather have it crash immediately than write an invalid file, because I might not even notice the file being invalid until it's too late.

I think this is a proper example: If you have to choose between `panic!` or "corrupted file", it make sense to panic. This is a good reason to return a proper error (IMHO). In your example "it will do what the Affinity Photo developpers choose" (unwrap/panic! or another handling/inform the user).

> the whole crate might have to be scanned to find similar assert statements.

I thought I already did that but I just realize my PyCharm is buggy and it can't find anything outside already opened (and so, indexed) files.

This PR has no sense if they are still `assert!` flying around. I thought there were only few asserts an thought it could be interesting to remove them for clean API, but I realize this PyCharm "indexing bug" just now. :(

I don't think it's a good idea to merge this half baked PR, because the point was to remove all asserts and it actually don't do that.

No, don't modify anything, this whole PR is just non sense. The discussion make sense, but what this PR do (changing 2 asserts and leaving the other) has no sense.

Sorry for this. :(

johannesvollmer:
Don't be sad. This pr is an important first step. There is no value lost by realizing that more has to be done later. We can merge it right now, and have one problem less in the future, or we can leave it as-is until we have the time to search for all assert statements. What do you think?

Narann:
Thanks. I think we should close it and open a ticket about "assert policy" and point to this PR (I will do it).

PR #114, opened 2021-04-18
prepare lossy compression methods

johannesvollmer:
draft. 
- [x] test this new algorithm using the existing lossy method (f32 pxr24)

@Narann

### DIFF
--- a/src/block/mod.rs
+++ b/src/block/mod.rs
@@ -515,15 +515,16 @@ impl UncompressedBlock {
         let absolute_indices = header.get_absolute_block_pixel_coordinates(tile_coordinates)?;
         absolute_indices.validate(Some(header.layer_size))?;
 
-        debug_assert_eq!(
+        if !header.compression.may_loose_data() { debug_assert_eq!(
             &header.compression.decompress_image_section(
                 header,
                 header.compression.compress_image_section(header, data.clone(), absolute_indices)?,
                 absolute_indices,
                 true
             ).unwrap(),
-            &data, "compression method not round trippin'"
-        );
+            &data,
+            "compression method not round trippin'"
+        ); }
 
         let compressed_data = header.compression.compress_image_section(header, data, absolute_indices)?;
 
--- a/src/compression/mod.rs
+++ b/src/compression/mod.rs
@@ -67,6 +67,7 @@ pub enum Compression {
     PIZ,
 
     /// Like `ZIP1`, but reduces precision of `f32` images to `f24`.
+    /// Therefore, this is lossless compression for `f16` and `u32` data, lossy compression for `f32` data.
     /// This produces really small image files. Only supported for flat images, not for deep data.
     // After reducing 32-bit floating-point data to 24 bits by rounding (while leaving 16-bit
     // floating-point data unchanged), differences between horizontally adjacent pixels
@@ -260,10 +261,32 @@ impl Compression {
     pub fn supports_deep_data(self) -> bool {
         use self::Compression::*;
         match self {
-            Uncompressed | RLE | ZIP1 | ZIP16 => true,
+            Uncompressed | RLE | ZIP1 /* | ZIP16*/ => true,
             _ => false,
         }
     }
+
+    /// Most compression methods will reconstruct the exact pixel bytes,
+    /// but some might throw away unimportant data for specific types of samples.
+    pub fn is_lossless_for(self, sample_type: SampleType) -> bool {
+        use self::Compression::*;
+        match self {
+            PXR24 => sample_type != SampleType::F32, // pxr reduces f32 to f24
+            Uncompressed | RLE | ZIP1 | ZIP16 | PIZ => true,
+            B44 | B44A | DWAB | DWAA(_) => false,
+        }
+    }
+
+    /// Most compression methods will reconstruct the exact pixel bytes,
+    /// but some might throw away unimportant data in some cases.
+    pub fn may_loose_data(self) -> bool {
+        use self::Compression::*;
+        match self {
+            Uncompressed | RLE | ZIP1 | ZIP16 | PIZ => true,
+            PXR24 | B44 | B44A | DWAB | DWAA(_) => false,
+        }
+    }
+
 }
 
 // see https://github.com/AcademySoftwareFoundation/openexr/blob/6a9f8af6e89547bcd370ae3cec2b12849eee0b54/OpenEXR/IlmImf/ImfMisc.cpp#L1456-L1541
--- a/src/compression/mod.rs
+++ b/src/compression/mod.rs
@@ -272,8 +272,9 @@ impl Compression {
         use self::Compression::*;
         match self {
             PXR24 => sample_type != SampleType::F32, // pxr reduces f32 to f24
+            B44 | B44A => sample_type != SampleType::F16, // b44 only compresses f16 values, others are left uncompressed
             Uncompressed | RLE | ZIP1 | ZIP16 | PIZ => true,
-            B44 | B44A | DWAB | DWAA(_) => false,
+            DWAB | DWAA(_) => false,
         }
     }
 
@@ -284,6 +285,16 @@ impl Compression {
         match self {
             Uncompressed | RLE | ZIP1 | ZIP16 | PIZ => false,
             PXR24 | B44 | B44A | DWAB | DWAA(_) => true,
+        }
+    }
+
+    /// Most compression methods will reconstruct the exact pixel bytes,
+    /// but some might replace NaN with zeroes.
+    pub fn supports_nan(self) -> bool {
+        use self::Compression::*;
+        match self {
+            B44 | B44A | DWAB | DWAA(_) => false, // TODO dwa might support it?
+            _ => true
         }
     }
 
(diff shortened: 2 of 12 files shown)


src/compression/mod.rs
```diff
@@ -260,10 +261,32 @@ impl Compression {
     pub fn supports_deep_data(self) -> bool {
         use self::Compression::*;
         match self {
-            Uncompressed | RLE | ZIP1 | ZIP16 => true,
+            Uncompressed | RLE | ZIP1 /* | ZIP16*/ => true,
```
Narann, line 264: Why commenting ZIP16?

src/compression/mod.rs
```diff
@@ -260,10 +261,32 @@ impl Compression {
...
     }
+
+    /// Most compression methods will reconstruct the exact pixel bytes,
+    /// but some might throw away unimportant data for specific types of samples.
+    pub fn is_lossless_for(self, sample_type: SampleType) -> bool {
+        use self::Compression::*;
+        match self {
+            PXR24 => sample_type != SampleType::F32, // pxr reduces f32 to f24
+            Uncompressed | RLE | ZIP1 | ZIP16 | PIZ => true,
+            B44 | B44A | DWAB | DWAA(_) => false,
```
Narann, line 276: From my understanding of `B44(A)` compression, only `f16` is compressed. The rest is raw copied.

tests/dev.rs
```diff
@@ -127,6 +127,5 @@ pub fn test_roundtrip() {
 
     println!("...read 2 successfull");
 
-    assert!(!image.contains_nan_pixels() && !image2.contains_nan_pixels());
-    assert_eq!(image, image2);
+    assert!(image.similar_to_lossy(&image2, 0.05));
```
Narann, line 130: Juste to be sure: Are you using `similar_to_lossy()` also for lossless compression?

Narann:
Looks good to me.

src/compression/mod.rs
johannesvollmer: Did you have to read the B44 code to find this out, or can I find it somewhere in the C++ comments? :)

src/compression/mod.rs
johannesvollmer: I have observed that the [C++ code actually does not support ZIP16](https://github.com/AcademySoftwareFoundation/openexr/blob/e46ac0a70cad0d84f8e8be743a9b2d633056da5b/src/lib/OpenEXR/ImfCompressor.cpp#L99-L109) 

(And I couldn't stop myself from fixing right in that moment, it in this PR, because I would otherwise forget to fix it hehe)

johannesvollmer: It's commented because ..eh i don't know... I guess I hoped it might be possible anyways haha. But we can also simply remove the commented ZIP16

tests/dev.rs
johannesvollmer: It's used everywhere, where an image is compared in the tests. I know this because I replaced all usages of `ContainsNaN` which was used in every test previously. These tests run for all exr images in the repository, including some B44 images. We can always add more images to the test folder to increase coverage.

johannesvollmer:
Great. As a surprise, I implemented some unit tests, so now there is more code to review :D

Please forgive me that I also made rather trivial changes to the API of the crate in the process, when I realized some parts might be missing. I felt very productive doing it and I really wanted to avoid forgetting to make these changes later.


- fix similar_to_lossy implementations to also compare the meta data data instead of only pixel data
- add rgb_channels (parallel to rgba_channels) when reading channels without having to deal with any alpha
- use non-recursive (tuple) types when creating SpecificChannels using the `rgb` or `rgba` constructors
- rename from_encoded_layer to from_encoded_channels

.github/workflows/rust.yml
```diff
@@ -2,9 +2,9 @@ name: Rust
 
 on:
   push:
-    branches: [ master ]
+    branches: [ master, dev ]
```
johannesvollmer, line 5: Yes, the automated github checks finally run! I wonder why this was necessary, as these changes should already be in the `dev` branch, so it should also already be in the other two pull requests

tests/dev.rs
johannesvollmer: However, I assume most images will not change. Because the image files in the repository have already been compressed with lossy methods, opening the images and compressing them again will likely not loose any additional data. We still have the tests from your PR though, where an image will be opened, then the compression method will be set to B44, and then it will be compressed. So I see no problem considering this.

src/block/mod.rs
```diff
@@ -515,15 +515,16 @@ impl UncompressedBlock {
         let absolute_indices = header.get_absolute_block_pixel_coordinates(tile_coordinates)?;
         absolute_indices.validate(Some(header.layer_size))?;
 
-        debug_assert_eq!(
+        if !header.compression.may_loose_data() { debug_assert_eq!(
```
johannesvollmer, line 518: This cannot use `similar_to_lossy`, because this code does not know whether these bytes are f16, f32, u32, or any combination of these. Being aware of that would be very complicated for a simple check.

src/compression/mod.rs
Narann: I find it while translating code. Here is the real code:

compress: https://github.com/AcademySoftwareFoundation/openexr/blob/master/src/lib/OpenEXR/ImfB44Compressor.cpp#L746
decompress: https://github.com/AcademySoftwareFoundation/openexr/blob/master/src/lib/OpenEXR/ImfB44Compressor.cpp#L888

It make sense as B44 is a hand crafted compression of f16 and seems to have no effect on u32/f32.

That's my understanding of the thing.

tests/dev.rs
Narann: I was asking because I think it can be a good thing to keep a strong equal comparison for non destructive compression.

src/block/mod.rs
Narann: Make sense.

tests/dev.rs
johannesvollmer: I think you're right, nice. Does this mean that f32 data will be completely uncompressed in the file?! 

Also, why is it necessary to copy these channels twice (in to tmp, tmp to out)? We could save one copy and some memory there (just an idea for later, not now).

src/compression/mod.rs
johannesvollmer: I think you're right, nice. Does this mean that f32 data will be completely uncompressed in the file?!

Also, why is it necessary to copy these channels twice (in to tmp, tmp to out)? We could save one copy and some memory there (just an idea for later, not now).

tests/dev.rs
johannesvollmer: Yes, I thought of that, but no design I came up with satisfied me. I think I have a better idea now, let me try

johannesvollmer: Pushed a new design implementing this idea

src/compression/mod.rs
Narann: > Edit: Yes it does. Oh no.

More than this: I suspect it's a specific feature to read some channels in real time and having other in full precision. Just a guess but from the original implementation, that's the behavior you could have.

> Also, why is it necessary to copy these channels twice (in to tmp, tmp to out)? We could save one copy and some memory there (just an idea for later, not now).

Yes ! That's the part of the code I need to review. That was my original goal and have difficulties with u8 and u16 array copy so I decided to be full u16 and fall to u8 in a big copy to the end, just to get the "B44 compression working", but its defenetly something to do. Once again, I need more time to mess with slices/types.

src/meta/header.rs
```diff
@@ -973,8 +975,14 @@ impl Header {
...
 
+        let dwa_level = match self.compression {
+            attribute::Compression::DWAA(level) => level,
+            attribute::Compression::DWAB(level) => level,
+            _ => None,
+        };
+
         // write compression parameters as attribute
-        if let attribute::Compression::DWAA(Some(level)) = self.compression {
+        if let Some(level) = dwa_level {
```
Narann, line 985: Why not:

```
// DWA writes compression parameters as attribute.
match self.compression {
  attribute::Compression::DWAA(level) |
  attribute::Compression::DWAB(level) => attribute::write(DWA_COMPRESSION_LEVEL, &F32(level), write)?;
  _ => (),
};
```
Just an idea.

src/compression/mod.rs
johannesvollmer: Ok

src/meta/header.rs
johannesvollmer: Nice.

johannesvollmer:
If there are no objections left @Narann, then we can merge this, and you can have pull tests into your B44 PR :)

All B44 tests automatically run with lossy precision. I guess most B44 images won't work yet, because they often also use luma/chroma and subsampling (which, by the way, we can finally implement once B44 is here). 

If you want to use it in your own tests, the latest design works like this:
1. `use crate::image::validate_results::ValidateImageResult;`
2. `let is_eq = original_image.validate_image_result(&lossy_image, 0.1);` (order matters)  
    The threshold is only used for images that use lossy compression methods

Narann:
OK, look good to me.

PR #122, opened 2021-05-03
Properly implement TimeCode

johannesvollmer:
(why does this contain changes from the dev branch??? I don't remember touching huffman.rs and piz.rs in this branch... anyways).

@Narann Does this contain all the functionality you expect from the `TimeCode` class?


src/meta/attribute.rs
```diff
@@ -1062,23 +1090,159 @@ impl ChannelList {
...
+                tv60_time_and_flags.get_bit(31),
+            ],
+
+            binary_groups: Self::unpack_user_data_from_u32(user_data)
+        }
+    }
+
+    /// Pack the SMPTE time code into a u32 value, according to TV50 packing.
+    /// This encoding does not support the `drop_frame` flag, it will be lost.
+    /// Panics for invalid values in the time code.
```
johannesvollmer, line 1155: Doesn't anymore

src/meta/attribute.rs
```diff
@@ -1062,23 +1090,159 @@ impl ChannelList {
...
+            ],
+
+            .. Self::from_tv60_time(tv50_time_and_flags, user_data)
+        }
+    }
+
+
+    /// Pack the SMPTE time code into a u32 value, according to FILM24 packing.
+    /// This encoding does not support the `drop_frame` and `color_frame` flags, they will be lost.
+    /// Panics for invalid values in the time code.
```
johannesvollmer, line 1189: Invalid documentation here also

src/meta/attribute.rs
```diff
@@ -1062,23 +1090,159 @@ impl ChannelList {
...
 impl TimeCode {
 
     /// Number of bytes this would consume in an exr file.
     pub const BYTE_SIZE: usize = 2 * u32::BYTE_SIZE;
 
-    /// Without validation, write this instance to the byte stream.
+    /// Returns an error if this time code is considered invalid.
+    pub fn validate(&self, strict: bool) -> UnitResult {
+        if strict {
+            if self.frame > 29 { Err(Error::invalid("time code frame larger than 29")) }
```
Narann, line 1102: > `Error::invalid("time code frame larger than 29")`

I'm not sure 29 is the limit: https://github.com/AcademySoftwareFoundation/openexr/blob/master/src/lib/OpenEXR/ImfTimeCode.cpp#L212

src/meta/attribute.rs
```diff
@@ -1062,23 +1090,159 @@ impl ChannelList {
...
+            else if self.seconds > 59 { Err(Error::invalid("time code seconds larger than 59")) }
+            else if self.minutes > 59 { Err(Error::invalid("time code minutes larger than 59")) }
+            else if self.hours > 23 { Err(Error::invalid("time code hours larger than 23")) }
+            else if self.binary_groups.iter().any(|&group| group > 4) {
+                Err(Error::invalid("time code binary group value too large for 3 bits"))
+            }
```
Narann, line 1108: Same here. I suspect the limit is 8: https://github.com/AcademySoftwareFoundation/openexr/blob/master/src/lib/OpenEXR/ImfTimeCode.cpp#L309

src/meta/attribute.rs
```diff
@@ -1062,23 +1090,159 @@ impl ChannelList {
...
+    }
+
+    // Unpack the encoded u32 user data to an array of bytes, each byte having a value from 0 to 4.
+    fn unpack_user_data_from_u32(user_data: u32) -> [u8; 8] {
+        (0..8).map(|group_index| user_data.get_bits(Self::user_data_bit_indices(group_index)) as u8)
+            .collect::<SmallVec<[u8;8]>>().into_inner().expect("array index bug")
+    }
+
+
+    /// Without validation, write this instance to the byte stream, without validation.
```
Narann, line 1233: "Without validation" written twice.

src/meta/attribute.rs
```diff
@@ -1062,23 +1090,159 @@ impl ChannelList {
...
     pub fn write<W: Write>(&self, write: &mut W) -> UnitResult {
-        self.time_and_flags.write(write)?;
-        self.user_data.write(write)?;
+        debug_assert!(self.validate(true).is_ok());
+        self.pack_time_as_tv60_u32()?.write(write)?;
```
Narann, line 1236: I wrote a long message before realizing the official implementation store timecode as TV60 internally (and so, assume this format), and TV50 and FILM24 are just bit reorganization.

Can I suggest to add a comment like "In official implementation, internal representation of time_and_flags is TV60, this is also how it's written in EXR file".

This is obvious on official implementation because timecode is not stored as multiple fields, but a single int. Here we choose the struct+field approach, so a reminder of the serialized format would make things clear: It's how the official implementation store it.

Narann:
Looks good to me, good work! :)

I think we should clear the `dev` branch modification before merging.

johannesvollmer:
Thanks for the feedback! Good catches.

Also, I completely forgot to implement the `bcd` conversion functions which means every bit lookup is wrong. I should probably create some test data with the original library and write a real test using that data...

src/meta/attribute.rs
johannesvollmer: I think the group parameter is an index, to identify one of 8 groups. 

The value uses three bits. So 2^3=8 should be the max. But when I tried to use 8 as max value, the set_bit function panicked because the value was too large, so I concluded that the limit must be 4. But it could also be 7, now that I think about it. 

Shouldn't 8 work for 3 bits?
1. `..000`
1. `..001` 
2. `..010`
3. `..011`
4. `..100`
5. `..101`
6. `..110`
7. `..111`
8. 
So 8 should not panic, right?

src/meta/attribute.rs
johannesvollmer: Haha :D yes sounds reasonable, I'll do that

johannesvollmer:
Major bug in my code: The C++ code uses inclusive indices, which is not common in Rust. So _bit range `0 - 5` from C++ should actually be `0..6` in Rust_, but it currently is `0..5`, which is one bit less.

src/meta/attribute.rs
johannesvollmer: 15, because I had an indices bug, and this field actually has 4 bits available instead of only 3

src/meta/attribute.rs
johannesvollmer: I have a problem with this: The value has 6 bits, which allows for 64 values. But the value is converted to a little bit more inefficient data format, "bcd" (binary encoded decimal), which would require more than 6 bits if the range is 0-59. I can't find where the bug is, could you have a quick closer look at this?

johannesvollmer:
Fixed the index bug, but there is a test failing: The frame values, after being converted to BCD format, are too large for 6 bits.
image

johannesvollmer:
Settled on using 29 as the limit for the `frame` field, as suggested in [external sources](http://www.philrees.co.uk/articles/timecode.htm).

src/meta/attribute.rs
johannesvollmer: In timecode.h they said 29 is the limit, and other websites say that too

johannesvollmer:
> I think we should clear the dev branch modification before merging.

That would be correct, but I would push these strange changes it into dev anyways. So because I'm lazy I'll just include these here, even if this is not a very clean procedure.

PR #138, opened 2021-06-02
fix on-progress type parameter

johannesvollmer:
fix on-progress type parameter


johannesvollmer:
and prepare new release

PR #147, opened 2022-01-03
Channel groups

johannesvollmer:


src/image/channel_groups.rs
```diff
@@ -1,152 +1,237 @@
...
+    fn try_map<T>(self, mut mapper: impl FnMut(Channels) -> Result<T>) -> Result<Groups<T>> {
+        let channels = match self.own_channels {
+            Some(channels) => Some(mapper(channels)?),
+            None => None,
+        };
+
+        let new_child_groups = HashMap::with_capacity(self.child_groups.len());
+        let child_groups = self.child_groups.into_iter()
+            .map(|(name, child)| Ok((name, child.try_map(&mut mapper)?)))
+            .try_fold(
```
johannesvollmer, line 120: https://doc.rust-lang.org/rust-by-example/error/iter_result.html

src/image/channel_groups.rs
```diff
@@ -155,43 +240,26 @@ impl<'slf, ChannelGroup> WritableChannels<'slf> for ChannelGroups<ChannelGroup>
...
-
-            // read from specific channel into total byte block
-            // this assumes that the lines in the callback are iterated in strictly increasing order
-            // because each channel reader is consumed
-            channel_reader.read_exact(line.value)
-                .expect("collecting grouped channel byte block failed");
-        })
+impl<Channels> ChannelsWriter for GroupChannelsWriter<Channels> where Channels: ChannelsWriter {
+    fn extract_uncompressed_block(&self, header: &Header, block_index: BlockIndex, output_block_data: &mut [u8]) {
+        for channels_group in &self.all_channel_groups {
```
johannesvollmer, line 260: Each subgroup will write to the same few first bytes, overwriting each other. Incorporate some kind of offset?

src/image/channel_groups.rs
```diff
@@ -155,43 +205,42 @@ impl<'slf, ChannelGroup> WritableChannels<'slf> for ChannelGroups<ChannelGroup>
...
-                    .expect("channels not found bug");
-
-                channels.create_writer(header) // channel_info.name.clone()
-            })
-            .collect();
-
-        GroupChannelsWriter { channels_list: channels }
+        GroupChannelsWriter {
+            all_channel_groups: self.all_channel_groups()
+                .map(|channel_group: &ChannelGroup| panic!("this uses relative names but expects absolute names, and all will write first byte")/*channel_group.create_channel_group_writer(header)*/)
```
johannesvollmer, line 213: `SpecificChannels` currently does a lookup based on the channel name, in order to find the byte offset, for each channel. This cannot work as the meta data channels have absolute names while the `SpecificChannels` don't

src/image/channel_groups.rs
```diff
@@ -201,67 +250,74 @@ struct ReadChannelGroups<ReadChannelGroup> {
...
 
-            // FIXME this is not the original order indexed_channels.len() - 1
-            indexed_channels[]
-        });
+                let group_selected_channel_indices = ChannelMask::only(group_selected_channel_indices);
 
-        Ok(ChannelGroupsReader {
-            channels: channel_groups,
-            indexed_channels,
+                let reader = self.read_channels.create_channels_reader(header, &group_selected_channel_indices);
```
johannesvollmer, line 285: At this point, meta data contains fully qualified names, but the specific channels and any channels must work with relative local names. Pass some kind of string map instead of only indices?

Issue #153, opened 2022-04-19
Krita 4.4.5 and RenderDoc 1.18 cannot open a trivial EXR file generated by the library example


curldivergence:
# Description
Hi, I've tried running a couple of examples, including the first one from README, but unfortunately, I'm unable to view the generated files neither with Krita nor with RenderDoc. The error messages I'm facing look like this (respectively):

![image](https://user-images.githubusercontent.com/1927180/164067068-a704c1de-7cb0-4e03-9efe-ade9ca943718.png)
![image](https://user-images.githubusercontent.com/1927180/164067132-7a5161ba-f014-4419-b7d5-c60c7fb5b1cb.png)

Is there any chance this is a known issue, or maybe I'm doing something wrong? :)

Thank you!

# Reproducing
1. Run the trivial example from README:
```rust
fn main() {
    // write a file with 16-bit alpha and 32-bit color precision
    exr::prelude::write_rgba_file(
        "tests/images/out/minimal_rgb.exr",
        2048, 2048, // write an image with 2048x2048 pixels
        |x,y| ( // generate (or lookup in your own image) an f32 rgb color for each of the 2048x2048 pixels
                x as f32 / 2048.0, // red
                y as f32 / 2048.0, // green
                1.0 - (y as f32 / 2048.0), // blue
                f16::from_f32(0.8) // 16-bit alpha
        )
    ).unwrap();
}
```
2. Try to open the generated .exr with Krita or RenderDoc
3. Observe the error 

# Environment
- Operating System: Windows 11
- CPU Model: Intel(R) Core(TM) i7-9700F
- EXRS Version used: 1.4.2

johannesvollmer:
Hi, thanks for taking the time. It's not a known issue. We only know that Affinity Designer and blender can open the images. Let's have a look at Krita and RenderDoc

johannesvollmer:
There is a few things we can try. For example, changing the compression method, or the difference between rgb and rgba, or a custom channel set

johannesvollmer:
Some observations:
- Exported exrs rgba image can be opened with affinity
- Exporting that file from Affinity and opening in Krita works only if I remove the alpha channel on Affinity (otherwise it is only white, maybe Krita or Affinity got confused by the channel naming)
- Exporting that image from Krita and printing the meta data reveals no special or suspicious meta data (not any at all, basically)

curldivergence:
[...]

PR #154, opened 2022-04-24
improve examples & (api-breaking) rename some constant

johannesvollmer:
& fix duplicate attribute validation bug


johannesvollmer:
the example about cropping still uses some potentially non-existent file

johannesvollmer:
Fixing #153

johannesvollmer:
Todo: make mip map example run, and use it in the example where pngs are extracted

Issue #165, opened 2022-09-09
Multiply with overflow


anfedotoff:
Hi!
We were using our tool Sydr to find some integer overflows and we found this at [line](https://github.com/johannesvollmer/exrs/blob/master/src/meta/attribute.rs#L1503): 
`thread '<unnamed>' panicked at 'attempt to multiply with overflow', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/exr-1.5.0/src/meta/attribute.rs:1503:45`

Technically it is a true positive, but allocation is safe. Maybe after values `width` and `height` will be used and something can go wrong?

johannesvollmer:
Thanks for the efforts! :) Do you have any more information? The input data, maybe? Or the code calling into the library?

johannesvollmer:
As this is just the preview image, this is only the case for a corrupted image file. It should not panic or overflow of course. We can simply make these multiplications safe, using checked multiplications. I'm sure this will already solve the issue.

anfedotoff:
> Thanks for the efforts! :) Do you have any more information? The input data, maybe? Or the code calling into the library?

:) We were fuzzing image-rs using this [harness](https://github.com/image-rs/image/blob/master/fuzz/fuzzers/fuzzer_script_guess.rs).  I attach the input file
[crash-sydr_8e195a006d8b5f59f3f0de7cb7e2a0f2e5b62843_int_overflow_0_unsigned.txt](https://github.com/johannesvollmer/exrs/files/9540017/crash-sydr_8e195a006d8b5f59f3f0de7cb7e2a0f2e5b62843_int_overflow_0_unsigned.txt)

anfedotoff:
> As this is just the preview image, it will never be this large, except for the case of corrupted image files. Of course, for these cases, it should never overflow. We can simply make these multiplications safe, using checked multiplications. I'm sure this will already solve the issue.

Good, I try to compose a PR than!

johannesvollmer:
That would be amazing!

PR #166, opened 2022-09-10
Fix multiply overflow at src/meta/attribute:1503

anfedotoff:
Fix for #165.

### DIFF
--- a/src/meta/attribute.rs
+++ b/src/meta/attribute.rs
@@ -1496,15 +1496,15 @@ impl Preview {
 
     /// Read the value without validating.
     pub fn read<R: Read>(read: &mut R) -> Result<Self> {
-        let components_per_pixel = 4;
         let width = u32::read(read)? as usize;
         let height = u32::read(read)? as usize;
 
-        if let Some(size) = width.checked_mul(height) {
-            if let Some(total_size) = size.checked_mul(components_per_pixel) {
+        if let Some(pixel_count) = width.checked_mul(height) {
+            // Multiply by the number of bytes per pixel.
+            if let Some(byte_count) = pixel_count.checked_mul(4) {
                 let pixel_data = i8::read_vec(
                     read,
-                    total_size,
+                    byte_count,
                     1024 * 1024 * 4,
                     None,
                     "preview attribute pixel count",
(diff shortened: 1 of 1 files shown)


src/meta/attribute.rs
```diff
@@ -1500,14 +1500,31 @@ impl Preview {
...
-            size: Vec2(width, height),
-            pixel_data,
-        };
+                return Ok(preview);
+            }
+        }
 
-        Ok(preview)
+        return Err(Error::invalid(
+                format!("Overflow while calculating image size \
```
johannesvollmer, line 1523: please mention in the error message that this is about the preview image Attribute, not the actual image

src/meta/attribute.rs
```diff
@@ -1500,14 +1500,31 @@ impl Preview {
...
+                (width: {}, height: {}, components_per_pixel: {}).",
```
johannesvollmer, line 1524: I would remove the components per pixel output, it's always 4. Or maybe rename to bytes per pixel. What do you think?

src/meta/attribute.rs
```diff
@@ -1500,14 +1500,31 @@ impl Preview {
         let width = u32::read(read)? as usize;
         let height = u32::read(read)? as usize;
 
-        let pixel_data = i8::read_vec(read, width * height * components_per_pixel, 1024*1024*4, None, "preview attribute pixel count")?;
+        if let Some(size) = width.checked_mul(height) {
+            if let Some(total_size) = size.checked_mul(components_per_pixel) {
```
johannesvollmer, line 1504: this variable (total_size) never had a nice name in the first place, could you rename it to `byte_count` while you're at it? Would be nice

src/meta/attribute.rs
```diff
@@ -1500,14 +1500,31 @@ impl Preview {
         let width = u32::read(read)? as usize;
         let height = u32::read(read)? as usize;
 
-        let pixel_data = i8::read_vec(read, width * height * components_per_pixel, 1024*1024*4, None, "preview attribute pixel count")?;
+        if let Some(size) = width.checked_mul(height) {
```
johannesvollmer, line 1503: Cool

johannesvollmer (changes requested):

src/meta/attribute.rs
anfedotoff: It's a good idea to remove. I think we could add a comment before multiplying by four, about what 4 means.

src/meta/attribute.rs
anfedotoff: `byte_count` sounds good. Maybe change `size`  to `pixel_count`?

johannesvollmer: Excellent, indeed

johannesvollmer:
Excellent work, thank you for your contribution! :)

johannesvollmer (approved):

PR #167, opened 2022-09-14
Enforce decompressed size limits

Shnatsel:
...now that I've [fixed](https://github.com/Frommi/miniz_oxide/pull/121) the limit enforcement in miniz_oxide, use the `*_with_limit` variants of decompression functions.

I've also bumped the miniz_oxide to the lowest fixed version. Using 0.6.x is probably fine but I didn't switch to 0.6.x out of caution - I'm not sure if any miniz_oxide types are exposed in the public interface or not.


johannesvollmer:
Excellent, thank you for taking the time! :)

johannesvollmer (approved):

PR #174, opened 2022-12-28
Add non-parallel RLE read benchmark

Shnatsel:
The parallel RLE read benchmark is highly unreliable, showing variance from 20 to 28 without any changes to the code.

This PR adds a single-threaded equivalent of that benchmark so that performance could be more accurately measured.

I also suggest removing the parallel version to avoid confusion in the future, but that's up to you and not included in this PR.

### DIFF
--- a/benches/read.rs
+++ b/benches/read.rs
@@ -60,6 +60,24 @@ fn read_single_image_rle_all_channels(bench: &mut Bencher) {
     })
 }
 
+/// Read without multi-core RLE decompression
+fn read_single_image_rle_non_parallel_all_channels(bench: &mut Bencher) {
+    bench.iter(||{
+        let path = "tests/images/valid/custom/crowskull/crow_rle.exr";
+
+        // copied from `read_all_flat_layers_from_file` and added `.non_parallel()`
+        let image = exr::prelude::read()
+            .no_deep_data()
+            .largest_resolution_level()
+            .all_channels()
+            .all_layers()
+            .all_attributes()
+            .non_parallel()
+            .from_file(path).unwrap();
+        bencher::black_box(image);
+    })
+}
+
 /// Read without multi-core ZIP decompression
 fn read_single_image_non_parallel_zips_rgba(bench: &mut Bencher) {
     bench.iter(||{
@@ -82,6 +100,7 @@ benchmark_group!(read,
     read_single_image_uncompressed_rgba,
     read_single_image_zips_rgba,
     read_single_image_rle_all_channels,
+    read_single_image_rle_non_parallel_all_channels,
     read_single_image_non_parallel_zips_rgba
 );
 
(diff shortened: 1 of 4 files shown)


johannesvollmer:
good point. great addition, and I'll think about removing, or at least clarifying, the other benchmark :)

johannesvollmer:
for a more nuanced benchmark suite, we can also add another RLE benchmark with a different storage type. Right now I see only `all_channels`, I'd like to add a `rgba` variation too.

Shnatsel:
So far I've been using the examples for profiling, but there often isn't a good one for what I'm looking for, so I expect I will start using benchmarks instead.

RGBA would be a great addition if it's the common case.

johannesvollmer:
great, the examples should cover most use cases, and should represent real-world use cases. For profiling though, we might want to read from in-memory buffers only, such that the delay produced by reading from disk is not too prominent

benches/read.rs
```diff
@@ -60,6 +60,24 @@ fn read_single_image_rle_all_channels(bench: &mut Bencher) {
     })
 }
 
+/// Read without multi-core RLE decompression
+fn read_single_image_rle_non_parallel_all_channels(bench: &mut Bencher) {
+    bench.iter(||{
+        let path = "tests/images/valid/custom/crowskull/crow_rle.exr";
```
johannesvollmer, line 66: Is it desired that the file is loaded from disk in each iteration? To avoid unnecessary wait times, I would suggest first loading the bytes from memory, then afterwards decode the image from those in-memory file bytes. We should probably change all benchmarks to use this strategy. You can see it in line 26, in the function `read_single_image_uncompressed_from_buffer_rgba`. 

This also makes sure that file access does not intruduce noise into the benchmark

Shnatsel: Makes sense. The inputs also have to be `black_box`ed in this case. Also `std::hint::black_box` was stabilized recently, it's probably a good idea to use it.

More info black boxing: [Why my Rust benchmarks were wrong, or how to correctly use std::hint::black_box?](https://gendignoux.com/blog/2022/01/31/rust-benchmarks.html)

johannesvollmer: I woud appreciate if you did that for all of the benchmarks. Including the black boxing of the input :)

Shnatsel: I've added from-memory and from-disk benchmarks for comparison. Not sure if they're useful at all, but I guess they validate that the from-disk one doesn't do anything stupid.

I've also split reading with format conversion into a separate file and explicitly labeled f32_to_f32 and f32_to_f16 conversions, and also added an f32_to_u32 benchmark.

Shnatsel: > I would appreciate if you did that for all of the benchmarks. Including the black boxing of the input :)

I don't want to pile even more things into this PR. Let's get this merged and I'll do a benchmark refactoring PR separately.

johannesvollmer: Great work! Thanks :) I'm curious to see how much buffering will speed up the decoding

johannesvollmer (approved):

PR #173, opened 2022-12-28
Faster, more Rusty interleaving

Shnatsel:
Massive speedups on reads across the board, with a 2x improvement on the `read_single_image_rle_all_channels` benchmark.

**Benchmarks:**

Before:
```
test read_single_image_non_parallel_zips_rgba        ... bench: 196,783,469 ns/iter (+/- 172,721)
test read_single_image_rle_all_channels              ... bench:  41,137,521 ns/iter (+/- 3,633,097)
test read_single_image_uncompressed_from_buffer_rgba ... bench:  33,522,367 ns/iter (+/- 70,401)
test read_single_image_uncompressed_rgba             ... bench:  37,435,859 ns/iter (+/- 2,202,515)
test read_single_image_zips_rgba                     ... bench:  71,419,820 ns/iter (+/- 4,860,501)
```

After:
```
test read_single_image_non_parallel_zips_rgba        ... bench: 150,670,222 ns/iter (+/- 10,422,001)
test read_single_image_rle_all_channels              ... bench:  21,199,353 ns/iter (+/- 4,196,646)
test read_single_image_uncompressed_from_buffer_rgba ... bench:  32,094,137 ns/iter (+/- 180,900)
test read_single_image_uncompressed_rgba             ... bench:  36,061,562 ns/iter (+/- 61,285)
test read_single_image_zips_rgba                     ... bench:  51,157,416 ns/iter (+/- 7,589,099)
```

**Profiles** on the `0c_read_rgba` example:

Before: https://share.firefox.dev/3PUnd5n

After: https://share.firefox.dev/3G2nzlM

As you can see, interleaving is down from taking up 58% to 14% of the time.

Most of the remaining time in interleaving is spent allocating and zeroing the Vec. Reusing the allocation can unlock further gains, at the cost of having to explicitly pass the Vec to the function for reuse. I can open a separate PR for that if you like.

PR #175, opened 2022-12-28
Faster `differences_to_samples`

Shnatsel:
This function is responsible for 55% of execution time after #173; optimizations here have a large effect on performance.

This change leverages instruction-level parallelism to speed it up.

Benchmarks before (current master):
```
test read_single_image_non_parallel_zips_rgba        ... bench: 196,783,469 ns/iter (+/- 172,721)
test read_single_image_rle_all_channels              ... bench:  41,137,521 ns/iter (+/- 3,633,097)
test read_single_image_rle_non_parallel_all_channels ... bench:  98,932,909 ns/iter (+/- 2,856,674)
test read_single_image_uncompressed_from_buffer_rgba ... bench:  33,522,367 ns/iter (+/- 70,401)
test read_single_image_uncompressed_rgba             ... bench:  37,435,859 ns/iter (+/- 2,202,515)
test read_single_image_zips_rgba                     ... bench:  71,419,820 ns/iter (+/- 4,860,501)
```

Benchmarks after:

```
test read_single_image_non_parallel_zips_rgba        ... bench: 183,337,110 ns/iter (+/- 4,920,259)
test read_single_image_rle_all_channels              ... bench:  38,442,440 ns/iter (+/- 4,608,694)
test read_single_image_rle_non_parallel_all_channels ... bench:  81,506,752 ns/iter (+/- 796,542)
test read_single_image_uncompressed_from_buffer_rgba ... bench:  31,839,775 ns/iter (+/- 81,663)
test read_single_image_uncompressed_rgba             ... bench:  34,882,260 ns/iter (+/- 86,133)
test read_single_image_zips_rgba                     ... bench:  69,871,756 ns/iter (+/- 6,176,810)
```

The effect is more pronounced when combined with #173  (as in, this change makes up a larger percentage of execution time).

Tests pass, and they did fail on earlier versions where I messed them up slightly.

If this approach is OK with you, I can also try to optimize `samples_to_differences` in a similar fashion, although that's more difficult because it iterates in reverse and references future values instead of past values.

### DIFF
--- a/src/compression/mod.rs
+++ b/src/compression/mod.rs
@@ -409,10 +409,19 @@ mod optimize_bytes {
 
     /// Integrate over all differences to the previous value in order to reconstruct sample values.
     pub fn differences_to_samples(buffer: &mut [u8]) {
+        // The naive implementation is very simple:
+        //
+        // for index in 1..buffer.len() {
+        //    buffer[index] = (buffer[index - 1] as i32 + buffer[index] as i32 - 128) as u8;
+        // }
+        //
+        // But we process elements in pairs to take advantage of instruction-level parallelism.
+        // When computations within a pair do not depend on each other, they can be processed in parallel.
+        // Since this function is responsible for a very large chunk of execution time,
+        // this tweak alone improves decoding performance of RLE images by 20%.
         let mut previous = buffer[0] as i16;
-        // Process elements in pairs to take advantage of instruction-level parallelism,
-        // since this function is responsible for a very large chunk of execution time.
         for chunk in &mut buffer[1..].chunks_exact_mut(2) {
+            // no bounds checks here due to indices and chunk size being constant
             let diff0 = chunk[0] as i16;
             let diff1 = chunk[1] as i16;
             let sample0 = (previous + diff0 - 128) as u8;
--- a/src/compression/mod.rs
+++ b/src/compression/mod.rs
@@ -449,12 +449,38 @@ mod optimize_bytes {
         //     buffer[index] = (buffer[index] as i32 - buffer[index - 1] as i32 + 128) as u8;
         // }
 
-        for i in (2..buffer.len()).rev().step_by(2) {
-            let (sample0, sample1, next_sample) = (buffer[i] as i32, buffer[i-1] as i32, buffer[i-2] as i32);
-            let diff0 = (sample0 - sample1 + 128) as u8;
-            let diff1 = (sample1 - next_sample + 128) as u8;
-            buffer[i] = diff0;
-            buffer[i-1] = diff1;
+        if let Some(first) = buffer.get(0) {
+            let mut previous = *first as i16;
+            for chunk in &mut buffer[1..].chunks_exact_mut(8) {
+                // no bounds checks here due to indices and chunk size being constant
+                let sample0 = chunk[0] as i16;
+                let sample1 = chunk[1] as i16;
+                let sample2 = chunk[2] as i16;
+                let sample3 = chunk[3] as i16;
+                let sample4 = chunk[4] as i16;
+                let sample5 = chunk[5] as i16;
+                let sample6 = chunk[6] as i16;
+                let sample7 = chunk[7] as i16;
+                // these computations do not depend on each other, unlike in the naive version,
+                // so they can be executed by the CPU in parallel via instruction-level parallelism.
+                // Unlike in decoding, here we don't have to do more work when using parallelism,
+                // so we can do way more work in parallel than in decoding
+                chunk[0] = (sample0 - previous + 128) as u8;
+                chunk[1] = (sample1 - sample0 + 128) as u8;
+                chunk[2] = (sample2 - sample1 + 128) as u8;
+                chunk[3] = (sample3 - sample2 + 128) as u8;
+                chunk[4] = (sample4 - sample3 + 128) as u8;
+                chunk[5] = (sample5 - sample4 + 128) as u8;
+                chunk[6] = (sample6 - sample5 + 128) as u8;
+                chunk[7] = (sample7 - sample6 + 128) as u8;
+                previous = sample7;
+            }
+            // handle the remaining element at the end not processed by the loop over pairs, if present
+            for elem in &mut buffer[1..].chunks_exact_mut(8).into_remainder().iter_mut() {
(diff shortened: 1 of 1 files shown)


Shnatsel:
I've also optimized compression in a similar way, for modest but noticeable performance improvement across the board.

Before:
```
test write_nonparallel_zip1_to_buffered      ... bench: 481,441,257 ns/iter (+/- 22,410,241)
test write_parallel_any_channels_to_buffered ... bench:  67,985,884 ns/iter (+/- 18,264,908)
test write_parallel_zip16_to_buffered        ... bench: 171,812,583 ns/iter (+/- 8,102,075)
test write_parallel_zip1_to_buffered         ... bench: 142,734,713 ns/iter (+/- 7,649,419)
test write_uncompressed_to_buffered          ... bench:  25,297,916 ns/iter (+/- 187,416)
```

After:
```
test write_nonparallel_zip1_to_buffered      ... bench: 468,520,748 ns/iter (+/- 7,942,493)
test write_parallel_any_channels_to_buffered ... bench:  62,532,362 ns/iter (+/- 7,417,551)
test write_parallel_zip16_to_buffered        ... bench: 165,498,070 ns/iter (+/- 5,988,365)
test write_parallel_zip1_to_buffered         ... bench: 139,177,276 ns/iter (+/- 10,863,506)
test write_uncompressed_to_buffered          ... bench:  24,443,239 ns/iter (+/- 149,205)
```

If there were a single-threaded RLE write benchmark, it would show a big improvement.

src/compression/mod.rs
```diff
@@ -408,16 +408,81 @@ fn mod_p(x: i32, y: i32) -> i32 {
...
+            }
         }
     }
 
     /// Derive over all values in order to produce differences to the previous value.
     pub fn samples_to_differences(buffer: &mut [u8]){
-        for index in (1..buffer.len()).rev() {
-            buffer[index] = (buffer[index] as i32 - buffer[index - 1] as i32 + 128) as u8; // index unsafe but handled with care and unit-tested
+        // naive version:
+        // for index in (1..buffer.len()).rev() {
```
johannesvollmer, line 448: love that you added a comment with the core idea of the algorithm :)

Shnatsel: Yes. My big gripe with lots of C code is that it only includes the optimized algorithm without explaining what it actually does. I really wanted to avoid this here.

Shnatsel:
I've optimized `samples_to_differences` further. It's now 5x faster than the naive version, or 10x faster with `-C target-cpu=native`.

johannesvollmer:
awesome :)

johannesvollmer:
Merged your first PR to master. Now pulling that new master into this PR. You will need to update your local branch after this.

src/compression/mod.rs
```diff
@@ -408,16 +408,103 @@ fn mod_p(x: i32, y: i32) -> i32 {
...
+        if let Some(first) = buffer.get(0) {
+            let mut previous = *first as i16;
+            for chunk in &mut buffer[1..].chunks_exact_mut(2) {
+                // no bounds checks here due to indices and chunk size being constant
+                let diff0 = chunk[0] as i16;
+                let diff1 = chunk[1] as i16;
+                // these two computations do not depend on each other, unlike in the naive version,
+                // so they can be executed by the CPU in parallel via instruction-level parallelism
+                let sample0 = (previous + diff0 - 128) as u8;
+                let sample1 = (previous + diff0 + diff1 - 128 * 2) as u8;
```
johannesvollmer, line 431: very interesting approach, it surprises me that it's faster, as we're now doing the computation twice! how did you come up with this approach? is that a common technique?

src/compression/mod.rs
```diff
@@ -408,16 +408,103 @@ fn mod_p(x: i32, y: i32) -> i32 {
...
+                let sample7 = chunk[7] as i16;
+                let sample8 = chunk[8] as i16;
+                let sample9 = chunk[9] as i16;
+                let sample10 = chunk[10] as i16;
+                let sample11 = chunk[11] as i16;
+                let sample12 = chunk[12] as i16;
+                let sample13 = chunk[13] as i16;
+                let sample14 = chunk[14] as i16;
+                let sample15 = chunk[15] as i16;
+                // Unlike in decoding, computations in here are truly independent from each other,
```
johannesvollmer, line 478: awesome!

johannesvollmer (approved):

src/compression/mod.rs
Shnatsel: Avoiding data dependencies to leverage instruction-level parallelism is a common technique, yes.

I have tried doing more, and it's a very slight improvement for 3 elements on my machine and about the same on 4 elements. I felt doing 3 might penalize less sophisticated hardware like a Raspberry Pi and wouldn't be worth it.

After writing this, I checked what OpenEXR is doing here,, and it has an implementation using explicit SIMD for SSE. But it's a lot of instructions, and at a glance I'm not convinced that it's going to be any faster than this. Plus this library is safe code only, so no explicit SIMD until portable SIMD stabilizes.

This is essentially computing a prefix sum, which is a common and well-studied primitive (well, with a few tweaks - the +previous and -128 bits, but those are easy to vectorize). You can find literature on it if you're willing to dig around. It's probably going to be talking about either thread parallelism or explicit SIMD, though.

johannesvollmer: I never doubted that this is the best approach for the library, I have also thought about adding explicit or implicit SIMD from the start, it's great that you took the time! :) I was just curious, don't worry :)

Shnatsel: Oh, I'm not convinced that this is the best possible approach for the library. But it's nearly 2x faster than the current one, so I took it.

johannesvollmer: It's less code and less complexity than explicit SIMD, and it is safe. How is this not the best approach for exrs? :)

Shnatsel: Maybe there's a clever solution that relies on autovectorization that I'm not seeing. Or someone came up with a super clever algorithm that leverages instruction-level parallelism even more. Or a myriad of other possibilities.

johannesvollmer: I'd consider that to be your approach:) not saying it will never be improved, but that it's the right direction in my point of view

PR #176, opened 2022-12-29
Reuse buffers in interleaving for a small speedup

Shnatsel:
Builds upon #173, please review and merge that first.

Allocating memory is cheap, but initializing it is not. This change makes the interleaving function reuse the buffers it has allocated and initialized. On RLE image decoding this further brings down the time spent in interleaving from 15% to 5%.

### DIFF
--- a/src/compression/mod.rs
+++ b/src/compression/mod.rs
@@ -406,8 +406,6 @@ fn mod_p(x: i32, y: i32) -> i32 {
 
 /// A collection of functions used to prepare data for compression.
 mod optimize_bytes {
-    use crate::compression::SCRATCH_SPACE;
-
 
     /// Integrate over all differences to the previous value in order to reconstruct sample values.
     pub fn differences_to_samples(buffer: &mut [u8]) {
@@ -519,7 +517,7 @@ mod optimize_bytes {
         static SCRATCH_SPACE: Cell<Vec<u8>> = Cell::new(Vec::new());
     }
 
-    fn with_reused_buffer<F>(length: usize, func: F) where F: FnMut(&mut [u8]) {
+    fn with_reused_buffer<F>(length: usize, mut func: F) where F: FnMut(&mut [u8]) {
         SCRATCH_SPACE.with(|scratch_space| {
             // reuse a buffer if we've already initialized one
             let mut buffer = scratch_space.take();
--- a/src/compression/mod.rs
+++ b/src/compression/mod.rs
@@ -404,8 +404,19 @@ fn mod_p(x: i32, y: i32) -> i32 {
     x - y * div_p(x, y)
 }
 
+use std::cell::Cell;
+thread_local! {
+    // A buffer for reusing between invocations of interleaving and deinterleaving.
+    // Allocating memory is cheap, but zeroing or otherwise initializing it is not.
+    // Doing it hundreds of times (once per block) would be expensive.
+    // This optimization brings down the time spent in interleaving from 15% to 5%.
+    static SCRATCH_SPACE: Cell<Vec<u8>> = Cell::new(Vec::new());
+}
+
 /// A collection of functions used to prepare data for compression.
 mod optimize_bytes {
+    use crate::compression::SCRATCH_SPACE;
+
 
     /// Integrate over all differences to the previous value in order to reconstruct sample values.
     pub fn differences_to_samples(buffer: &mut [u8]){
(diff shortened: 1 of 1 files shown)


johannesvollmer:
I appreciate your input and think it's an acceptable solution. There is one thing I want to try first though, before merging: Would it be possible to swap the values within the slice without allocating any other memory at all?
I think it might be possible to only swap values from one index to another index, or maybe we need to track four indices? If that would be possible, I'd like to compare the performance and the complexity of the algorithm. Maybe you have an idea? Also I don't want to have a situation where we merge your code and then I replace it immediately.

Shnatsel:
As far as I understand the algorithm, as we swap elements in place, we need to keep more and more of them buffered somewhere. It's impossible to implement interleaving of unbounded size without scratch space; or perhaps I simply haven't figured out how.

There is a solution using half the memory, but the scratch space on bechmarks is just 30kB, and bringing the memory overhead down to 15kB is not worth the trouble of a more complex (and almost certainly slower) algorithm.

The memory copy at the end is surprisingly cheap - `copy_from_slice()` doesn't even show up on profile. With this function now only being responsible for 5% of execution time, optimizing it further is not really worth the effort.

johannesvollmer:
yes, after thinking about it, it most certainly won't be as performant to swap in-place. Are there any limitations to using statics in the library? As it is per thread, it should not have any observable side effects, right? Any problems on WASM or any other platform?

Shnatsel:
Indeed, buffer reuse should not be observable from the outside. It's an internal implementation detail.

I'm not aware of any limitations.  `Cell` is not thread-safe (it is `Send` but not `Sync`), but since it's used in a thread-local variable that shouldn't matter. There's no risk of runtime panics either because it's `Cell` and not `RefCell`. 

Platform support should not be an issue, I've used thread-local storage in WASM and even asm.js with no issues.

johannesvollmer:
thanks for the clarification! I'd like to merge all your code, but I'll not have access to my laptop for a few days or maybe even a week

johannesvollmer:
Merged your first PR to master. Now pulling that new master into this PR. You will need to update your local branch after this.

src/compression/mod.rs
```diff
@@ -404,8 +404,19 @@ fn mod_p(x: i32, y: i32) -> i32 {
     x - y * div_p(x, y)
 }
 
+use std::cell::Cell;
+thread_local! {
```
johannesvollmer, line 408: Have you thought about using `lazy_static`? I wouldn't mind bringing in that extra dependency for code simplification. Could we get rid of `Cell` here, possibly also simplifying the usage of the scratch buffer further down the line?

Also, can we make this more local by introducing

src/compression/mod.rs
```diff
@@ -510,74 +521,98 @@ mod optimize_bytes {
...
-        let mut interleaved = vec![0; separated.len()];
-        let (first_half, second_half) = separated
-            .split_at((separated.len() + 1) / 2);
+        SCRATCH_SPACE.with(|scratch_space| {
+            // reuse a buffer if we've already initialized one
+            let mut buffer = scratch_space.take();
+            if buffer.len() < separated.len() {
+                // efficiently create a zeroed Vec by requesting zeroed memory from the OS
+                buffer = vec![0u8; separated.len()];
+            }
```
johannesvollmer, line 530: Let's refactor this part into it's own function next to the declaration of the static variable. Not only is it roughly the same for both usages, but more importantly, it will separate the memory management from the two actual byte algorithms. This will simplify both algorithms. What do you think? 

Of course, only if we don't use `lazy_static` or if using it still requries some memory management.


Note: Including the finalizing statements of course
```
            // write out the results
            separated.copy_from_slice(&interleaved);
            // save the internal buffer for reuse
            scratch_space.set(buffer);
```

johannesvollmer (changes requested):

src/compression/mod.rs
Shnatsel: `lazy_static!` is global, while we need something thread-local. It also would require a `Mutex` to be mutable, which is more expensive than `Cell`, and no simpler to use.

src/compression/mod.rs
Shnatsel: Yes, it is faster. The alternative is a `memcpy()` for copying over old data plus `memset()` to expand the buffer. We don't need the old data, and requesting zeroed memory from the OS is faster than a `memset()`.

Shnatsel: But I understand this is done once per image. So it doesn't really matter if this is fast or not since it's done so rarely.

src/compression/mod.rs
johannesvollmer: oh right! perfect

src/compression/mod.rs
Shnatsel: I've factored out allocation reuse into its own function.

johannesvollmer (approved):
awesome! thanks :)

src/compression/mod.rs
johannesvollmer: great work :)

Issue #178, opened 2022-12-31
Pixel format conversions are slow


Shnatsel:
Right now format conversion invokes a callback on every pixel access, and is generally oriented towards per-pixel processing: 

https://github.com/johannesvollmer/exrs/blob/49fece069f9bc1e21b252fe0e98311867c38f5d3/src/image/read/specific_channels.rs#L279-L307

Note the `get_pixel: impl Fn(&mut FullPixel) -> &mut Sample` in the signature.

This prevents the compiler from vectorizing the code. There are vector instructions to turn batches of values of one type into another, but this interface prevents their use.

Lack of vectorization is especially bad for `f16` types, which do not exist natively on the CPU, and there are only vector instructions to convert to/from `f16`. See #177 for details - it's closed in favor of this one, but is still a problem.

For `f16` conversions not to be atrociously slow, they need to be batched together explicitly. The `half` crate provides functions converting slices to/from `f16`, see e.g. [`convert_from_f32_slice`](https://docs.rs/half/2.1.0/half/slice/trait.HalfFloatSliceExt.html#tymethod.convert_from_f32_slice).

For other types, straightforward conversion loops such as these will be automatically vectorized:

```rust
input.iter().zip(output.iter_mut()).for_each(|in_val, out_val| *out_val = in_val.into());
```
provided that `input` and `output` are both slices. You can also convert into a Vec:
```rust
let output: Vec<_> = input.iter().map(|in_val| in_val.into() ).collect();
```

johannesvollmer:
The code for conversion is pretty complex as it is. That's not good, of course, I'll explain why it is already rather complex: 

The user of the library can specify the type of pixels they expect by declaring a tuple type, for example `(f16, f32, f32, u32, f16, f16)`. The library will automatically find the channels with the specified names in the file, and extract only the desired channels from the image file. At the same time it will arrange them into a tuple of the requested form. Some values may be filled in if the channel was declared optional and is not present in the file. All of that is type safe, and as much as possible is done at compile time. 

[...]

PR #179, opened 2022-12-31
Use `zune-inflate` instead of `miniz_oxide` for decompression; it's faster

Shnatsel:
[zune-inflate](https://crates.io/crates/zune-inflate) is a new DEFLATE/Zlib library in 100% safe Rust that's faster than `miniz_oxide` and even `zlib-ng`.

Features:
 - 100% safe Rust (like `miniz_oxide`)
 - Can decompress any data compressed by `miniz_oxide` and `zlib_ng`, verified by me with a fuzzer
 - Fuzzed for 1 billion (with a B) executions on random data, also by me, no hangs/panics/OOMs remaining
 - Very fast (beats even `zlib-ng`)

Drawbacks:
 - Slightly less precise tracking of output limits

Benchmark before:

```
test read_single_image_non_parallel_zips_rgba        ... bench: 196,783,469 ns/iter (+/- 172,721)
```

Benchmark after:

```
test read_single_image_non_parallel_zips_rgba        ... bench: 185,493,451 ns/iter (+/- 1,794,656)
```

Not a huge difference by itself, because a lot of time is spent in other things I'm optimizing, but it does add up. Profile: https://share.firefox.dev/3QloVgt

This profile clearly shows time spent in things touched by #173, #175 and not-yet-PR'd #178, in addition to inflate.

With all my optimization PRs combined this benchmark is down to:

```
test read_single_image_non_parallel_zips_rgba        ... bench: 129,482,284 ns/iter (+/- 200,734)
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -626,7 +626,8 @@ checksum = "712e227841d057c1ee1cd2fb22fa7e5a5461ae8e48fa2ca79ec42cfc1931183f"
 [[package]]
 name = "zune-inflate"
 version = "0.2.0"
-source = "git+https://github.com/etemesi254/zune-image.git#3fd5558cf483f5c6b1455cbe9ab97fe92f2dd942"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b0bc976eac2cf2935926045b751d9bc7f1a268aa9ba0e11bef491dce239c7f26"
 dependencies = [
  "simd-adler32",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -144,6 +144,7 @@ dependencies = [
  "smallvec",
  "threadpool",
  "walkdir",
+ "zune-inflate",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


Shnatsel:
Nevermind, I think it needs more time to bake. I've re-run the roundtrip fuzz target and it seems to be broken again.

etemesi254:
Hi @johannesvollmer , author of zune-inflate


First of all, amazing , commendable and beautiful what you have here. 

But I would love your help to make zune-inflate at least faster for exr 

Specifically my concern and reference image isthe crow.exr   

First would you be kind enough to extract the zlib stream or at least show me how to do it from exr file

I tried for crow and got a 305b file, so i think I did something wrong :)


Also would like to know rhe characteristics lf a zlib stream in exr, i.e its typical size, whether it's preprocessed or not and whether this library decodes per line meaning it calls miniz-oxide repeatedly or at once

Would highly appreciate
Regards
Caleb Etemesi

johannesvollmer:
Hi! Thanks :) I'll drop the info in an issue in your repository, ok?

etemesi254:
Okay, works for me

Issue #172, opened 2022-12-27
Creating a SmallVec for every pixel is slow


Shnatsel:
I've been looking into the TODO in the README:

`TODO profile if smallvec is really an improvement!`

`SmallVec` is all over the place, so it's rather difficult to remove it and benchmark the before/after. I've used Rust's [instrumentation-based coverage](https://doc.rust-lang.org/stable/rustc/instrument-coverage.html) to find out how frequently a `SmallVec` is constructed. I've profiled the `7_crop_alpha_any_image` example. Here's the report in HTML (trimmed so that only covered lines are shown): [results_gt1.html.gz](https://github.com/johannesvollmer/exrs/files/10310525/results_gt1.html.gz)

I've found that an instance of this type is constructed for every pixel:

https://github.com/johannesvollmer/exrs/blob/49fece069f9bc1e21b252fe0e98311867c38f5d3/src/image/mod.rs#L493

Profiling shows that 55% of the execution time is spent just creating `SmallVec`s for every pixel in the crop function: https://share.firefox.dev/3WQyWUR

This will be even worse on images with more than 8 channels, where a heap allocation will be performed for every pixel.

It seems that the approach of creating a SmallVec for every pixel is fundamentally slow. A much more performant way to implement this would be returning a slice of the original data, without constructing an owned data structure.

Shnatsel:
The [`sample_vec_at`](https://github.com/johannesvollmer/exrs/blob/49fece069f9bc1e21b252fe0e98311867c38f5d3/src/image/mod.rs#L498-L501) function is to blame.

johannesvollmer:
thanks for investing the time! great observations :)

johannesvollmer:
the most important question is: is smallvec slower than a standard vec? because usually, slices are used where possible already. most of the time, smallvec is only used where sliced cannot be used unfortunately

Shnatsel:
Instead of creating a SmallVec for every pixel, you can have the caller provide a `&mut [u8]` to which the pixel will be written. This allows for a single allocation to be reused.

```
    pub fn output_sample_at(&self, position: Vec2<usize>, output_buffer: &mut [u8]) -> FlatSamplesPixel {
        output_buffer.copy_from_slice(self.samples_at(position))
    }
```

And the usage would look roughly like this:

```rust
// pre-allocate a buffer once; a Vec is faster in this case
let mut buffer = vec![0; bytes_per_pixel];
// iterate over all pixels in an image
[...]

PR #180, opened 2023-01-02
Implement size hints for `FlatSampleIterator`

Shnatsel:
This lets the caller know how much memory they should allocate.

Doesn't seem to do much to address #172, but may help some users since it's part of the public API.

### DIFF
--- a/src/image/mod.rs
+++ b/src/image/mod.rs
@@ -530,8 +530,14 @@ impl Iterator for FlatSampleIterator<'_> {
         }
         else { None }
     }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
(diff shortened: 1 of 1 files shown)


Shnatsel:
I've also implemented `nth()` so that the given channel can be looked up directly.

It's a terrible idea to use this iterator if you want to process a specific channel, but this impl makes it slightly less terrible.

(frankly it's a bad idea to use this iterator in the first place. Any processing that can be done on a per-channel basis should be - just get a slice of the channel and process it in chunks; this helps with cache utilization and memory bandwidth, and allows vectorization).

johannesvollmer (approved):

johannesvollmer (approved):

johannesvollmer:
thanks :) great addition

PR #181, opened 2023-01-02
Better read benchmarks

Shnatsel:
- Move reading from file into a buffer out of the benchmarking loop
- Use native `f32` instead of `f16` for benchmarking reads (conversion to `f16` is slow, which adds noise, and is profiled separately already)
- Add RLE RGBA benchmarks
- Ensure all benchmarks have a parallel and non-parallel version

### DIFF
--- a/benches/read.rs
+++ b/benches/read.rs
@@ -83,6 +83,46 @@ fn read_single_image_rle_non_parallel_all_channels(bench: &mut Bencher) {
     })
 }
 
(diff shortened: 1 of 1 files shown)


Shnatsel:
I'm not sure if the current zip benchmark is zip1 or zip16, but either way it would be nice to have both.

johannesvollmer (approved):

Issue #185, opened 2023-01-03
Benchmark against the C implementation of OpenEXR


Shnatsel:
## What can be improved or is missing?
Provide benchmarks comparing the performance of this crate to the [OpenEXR reference implementation](https://github.com/AcademySoftwareFoundation/openexr).
## Implementation Approach
The [`openexr`](https://docs.rs/openexr/latest/openexr) crate provides high-level, mostly safe bindings to the C implementation.

Shnatsel:
It would be interesting to have benchmarks on x86 as well as ARM.

I've tried running the benchmarks from #181 on a 16-core Ampere Altra machine from Google Cloud, and `exrs` in parallel mode absolutely rips :rocket: 

```
running 3 tests
test read_image_rgba_f32_to_f16 ... bench:  32,518,506 ns/iter (+/- 176,580)
test read_image_rgba_f32_to_f32 ... bench:  12,701,451 ns/iter (+/- 205,549)
test read_image_rgba_f32_to_u32 ... bench:  13,428,159 ns/iter (+/- 93,722)

test result: ok. 0 passed; 0 failed; 0 ignored; 3 measured

     Running benches/profiling.rs (target/release/deps/profiling-ddc84dc3d9a8e9fd)

running 2 tests
test read_single_image_all_channels             ... bench:  22,543,779 ns/iter (+/- 598,279)
test read_single_image_from_buffer_all_channels ... bench:  19,874,611 ns/iter (+/- 439,578)

test result: ok. 0 passed; 0 failed; 0 ignored; 2 measured

     Running benches/read.rs (target/release/deps/read-35e4db800494d5a6)

running 8 tests
test read_single_image_rle_all_channels               ... bench:  23,277,019 ns/iter (+/- 3,901,982)
test read_single_image_rle_non_parallel_all_channels  ... bench:  33,362,265 ns/iter (+/- 293,883)
test read_single_image_rle_non_parallel_rgba          ... bench:  36,240,049 ns/iter (+/- 247,417)
test read_single_image_rle_rgba                       ... bench:  16,579,403 ns/iter (+/- 301,560)
test read_single_image_uncompressed_non_parallel_rgba ... bench:  12,898,483 ns/iter (+/- 171,469)
test read_single_image_uncompressed_rgba              ... bench:  13,151,095 ns/iter (+/- 162,987)
test read_single_image_zips_non_parallel_rgba         ... bench: 137,174,659 ns/iter (+/- 417,130)
[...]

PR #191, opened 2023-01-07
batched f16 conversion

johannesvollmer:
(and also fix round up division missing documentation)

@Shnatsel will this approach work in terms of optimization? i had to add a few copy operations for technical reasons

to do:
- [x] unit tests 
- [x] before and after benchmark
- [x] add documentation about how to unlock all the optimizations (compiler flags and such)
- [x] use newest `half` dependency 
- [x] clean up and refactor
- [x] improve documentation of the functions


johannesvollmer:
before merging i want to have unit tests for that function and i want to clean it up, deduplicate the code, make it rusty

src/block/samples.rs
```diff
@@ -122,31 +124,77 @@ pub trait FromNativeSample: Sized + Copy + Default + 'static {
 
     /// Create this sample from a u32, trying to represent the same numerical value
     fn from_u32(value: u32) -> Self;
+
+    /// Convert all values from the slice into this type.
+    /// This function exists to allow the compiler to perform a vectorization optimization.
+    #[inline]
+    fn from_f16s(from: &[f16], to: &mut [Self]) {
```
Shnatsel, line 131: Auto-vectorization is not going to happen for `f16`. This should recommend overriding this method instead of referencing auto-vectorization.

johannesvollmer:
also i wonder whether the batch size of 4 can allow the compiler to optimize away all of the chunking logic in `HalfFloatSliceExt::convert_from_f32_slice`? Unfortunately we need to define a batch size, in order to avoid allocating a temporary buffer on the heap

src/image/read/specific_channels.rs
```diff
@@ -279,30 +280,85 @@ pub struct OptionalSampleReader<DefaultSample> {
...
+            mut bytes: impl Read,
+            mut out_samples: impl ExactSizeIterator<Item=&'t mut To>,
+            convert_slice: impl Fn(&[From], &mut [To])
+        ) where From: Data + Default + Copy, To: 't + Default + Copy
+        {
+            // using a batch size of 4
+            // because that's what `half` has vectorization for,
+            // and we want the compiler to
+            // optimize away all the logic in
+            // `HalfFloatSliceExt::convert_from_f32_slice`
```
Shnatsel, line 323: `half` supports converting 8 values at once in the latest git: https://github.com/starkat99/half-rs/issues/66 (not released to crates.io yet)

AVX-512 brings conversions of 16 values at once, which `half` also intends to support: https://github.com/starkat99/half-rs/issues/62

Also, right now `half` performs feature detection for each slice of 4 values, but the latest git performs it once per slice: https://github.com/starkat99/half-rs/commit/bc38c5b237e822e187cc1ac710b1099cc2faafd3
So bigger slice sizes are beneficial because the check only needs to be done once per the whole big slice instead of once per chunk of 4 or 8.

This needs to be balanced against not using too much CPU cache. I suggest the length of 16 for a start, since that's the minimum length that can take advantage of an AVX-512 instruction.

Shnatsel (commented):
The broad strokes look about right. 

I'll need to benchmark it and see if this actually improves performance, maybe also inspect the assembly to make sure these things are actually vectorized.

src/block/samples.rs
johannesvollmer: It is in fact overridden in the implementation. But sure, we can hint to that in this part of the documentation too.

src/image/read/specific_channels.rs
johannesvollmer: Can we somehow avoid having our own batching logic, and inside each batch still have the batching logic from `half`? I imagine it would be rather duplicate and lead to some performance problems

Shnatsel: That's actually a good thing! First we slice the input into batches of a size we choose that fits into the CPU cache, and `half` can then perform feature detection once per large slice (instead of once per small batch) and convert numbers in the most efficient way available in hardware, with a batch size we don't have to know or care about.

`chunks_exact_mut` that the `half` loop uses is cheap, and with a fixed-size slice (like we have) it will be cheaper still. Even the logic handling the remainder should be simply removed by the compiler, since the slice method in `half` is `#[inline]` and the batch size is constant.

Shnatsel:
Benchmarks: with `half = {git = "https://github.com/starkat99/half-rs", features = ["use-intrinsics"]}` in `Cargo.toml` I'm seeing the conversion time drop so much that it becomes unnoticeable, indistinguishable from the regular `f32` to `f32` codepath.

The `f32` to `u32` path doesn't get any faster, however. Maybe autovectorization doesn't kick in. I've tried with `-C target-cpu=native` for comparison, no difference.

Shnatsel:
ARM has native instructions for casting from `f32` to unsigned integers ([details](https://developer.arm.com/documentation/ddi0596/2021-12/SIMD-FP-Instructions/FCVTAU--vector---Floating-point-Convert-to-Unsigned-integer--rounding-to-nearest-with-ties-to-Away--vector--)). This would explain why I'm seeing [good results](https://github.com/johannesvollmer/exrs/issues/185#issuecomment-1369736550) for `f32` to `u32` conversion on ARM, even without this change.

I haven't found any mentions of native `f32` to `u32` casts on x86. ChatGPT mentioned some but I think it's making it up, and when I corrected it, it said this:

Снимок экрана от 2023-01-07 19-26-25

Shnatsel:
This also regresses the fallback path on half `2.2.0` but not on the latest git; we will need to switch to latest `half` so that we don't introduce regressions when intrinsics are not available.

Shnatsel:
Yeah, there seems to be no native conversion from `f32` to `u32` on x86_64 without AVX-512: https://stackoverflow.com/questions/41144668/how-to-efficiently-perform-double-int64-conversions-with-sse-avx

There is a conversion from `f32` to `i32`, but that would truncate large numbers that fit into `u32` but not `i32`, and therefore produce incorrect results (although `u32` cannot represent all `f32` values in the first place).

Why is `u32` is the chosen format, anyway?

Shnatsel (dismissed):
There's an outstanding nit about a comment and we'll need to bump to the latest `half` once the next release ships, otherwise looks good.

johannesvollmer:
Thanks for the feedback! Conversion from and to `u32` are only for completeness. I don't expect it to ever happen in the real world. We would not worry about inferior performance here at all. As you said, it's not even accurate for large numbers.

johannesvollmer:
>  I'm seeing the conversion time drop so much that it becomes unnoticeable, indistinguishable from the regular f32 to f32 codepath.

Awesome, didn't expect that much of a speed up!

johannesvollmer:
Do we have any regression concerning `f32 -> f32`?

src/image/read/specific_channels.rs
johannesvollmer: that's just what i hoped for, great :)

johannesvollmer:
> Why is u32 is the chosen format, anyway?

the purpose of `u32` samples is too assign different areas in the image unique IDs. The common case, rgba images, are either full `f32` or full `f16`.

Shnatsel:
> Do we have any regression concerning f32 -> f32?

Nope, it's the exact same on my machine. I guess the buffer does fit entirely into the L1 cache, it's not big.

johannesvollmer:
added a Todo list in the pr text

Shnatsel:
The necessary changes to `half` have shipped in version 2.2.1

johannesvollmer:
The code is not pretty, but it will be good enough for the moment. If there is no performance regression, I will merge it. @Shnatsel do you want to make sure it is still properly optimized after the refactorings?

Shnatsel:
You should see the difference on `cargo bench` yourself, but sure, I can take a look over the weekend.

johannesvollmer:
Yes, I was planning to see the benchmarks, sure :)

johannesvollmer:
Don't worry, take the time you need, does not have to be this weekend :)

johannesvollmer:
Strangely, on my machine, there is no performance gain whatsoever, rather within the margin of error, it is the same, if not worse. Note that this is the first time I benchmarked the code for this PR.

```
# no batch conversion
read_image_rgba_f32_to_f16 ... bench:  40,168,210 ns/iter (+/- 1,880,333)
read_image_rgba_f32_to_f32 ... bench:  19,554,980 ns/iter (+/- 2,367,993)
read_image_rgba_f32_to_u32 ... bench:  23,609,890 ns/iter (+/- 2,525,909)

# with batch conversion, dirty
read_image_rgba_f32_to_f16 ... bench:  45,213,290 ns/iter (+/- 2,912,926)
read_image_rgba_f32_to_f32 ... bench:  19,643,090 ns/iter (+/- 1,126,170)
read_image_rgba_f32_to_u32 ... bench:  26,500,130 ns/iter (+/- 2,043,967)

# with batch conversion, refactored
read_image_rgba_f32_to_f16 ... bench:  43,848,780 ns/iter (+/- 2,427,347)
read_image_rgba_f32_to_f32 ... bench:  19,653,510 ns/iter (+/- 2,083,299)
read_image_rgba_f32_to_u32 ... bench:  26,658,960 ns/iter (+/- 3,115,560)

```

johannesvollmer:
if there is something the user and me needs to do for good performance, this is the perfect time for some documentation

Shnatsel:
Yes, you need to enable the `use-intrinsics` feature in `half` for performance. It currently requires nightly, but will be stabilized in the upcoming Rust 1.68.

It's possible to make the implementation for 64-bit ARM to work on stable, but that's not yet merged into `half`: https://github.com/starkat99/half-rs/issues/63

Shnatsel:
Latest code by default:

```
test read_image_rgba_f32_to_f16 ... bench:  38,017,728 ns/iter (+/- 132,337)
```

With the `use-intrinsics` feature:

```
test read_image_rgba_f32_to_f16 ... bench:  14,419,196 ns/iter (+/- 146,359)
```

Also there's no need to set `-Ctarget-cpu=native`, since `half` performs runtime feature detection.

johannesvollmer:
I propose we introduce a feature flag to `exr`. The feature flag will enable `use-intrinsics` in `half`, but may be used for other purposes in the future. This way, people can still use the newest library with a slightly older old compiler. Maybe a flag of the name `nightly`? or "bleeding-edge"?

Shnatsel:
I suggest we simply wait for the next `half` release and switch to it once it's released. I understand it won't require any additional features.

I don't think exposing an `exr` feature that's only going to be relevant until Rust 1.69 ships is a good idea. Better document the workaround with depending on `half` and enabling the feature in it in the interim.

Shnatsel:
For context, `half` is going to do a major version bump with the stabilization of x86 intrinsics and the addition of ARM intrinsics. I don't think there's a way to select a major version of a library depending on a feature.

Shnatsel:
Speaking of, since it will take until Rust 1.69 to reap the benefits of this change on stable, it might be better to ship to make a new release without this PR. 

We can merge it immediately afterwards, and cut a new release once Rust 1.69 has shipped.

johannesvollmer:
Do I understand correctly that the newest Rust version is only required when we enable the `use-intrinsics` flag in `half`? In that case, if we leave the decision up to the user, it will be fine to just update the `half` dependency and let the people who have access to nightly Rust enable the flag.

Shnatsel:
Yes, you understand correctly - that is the behavior of the currently released 2.2.1 version.

A version requiring 1.69 will ship at a later date and will likely either enable the feature by default or remove it outright. It will also come with a major semver bump in `half`.

johannesvollmer:
aight, Rust 1.68 is here now!

> you need to enable the use-intrinsics feature in half for performance. It currently requires nightly, but will be stabilized in the upcoming Rust 1.68.

do we need a new release of `half` for the stable feature flag?

Shnatsel:
I believe stabilization of the required intrinsics got pushed back to 1.69, let me double-check that.

johannesvollmer:
yes, I think I didn't see any intrinsic in the changelog

Shnatsel:
So the stabilization PR is https://github.com/rust-lang/stdarch/pull/1366

It clearly didn't ship in 1.68

Shnatsel:
`half` 2.3.1 got rid of the feature flag and enabled the use of intrinsics by default, so no changes to `half` feature flags are needed on our side. This PR should be good to go, just needs unit tests.

src/block/samples.rs
```diff
@@ -122,31 +124,80 @@ pub trait FromNativeSample: Sized + Copy + Default + 'static {
 
     /// Create this sample from a u32, trying to represent the same numerical value
     fn from_u32(value: u32) -> Self;
+
+    /// Convert all values from the slice into this type.
+    /// This function exists to allow the compiler to perform a vectorization optimization.
+    /// Note that this default implementation will **not** be vectorized by the compiler automatically.
```
Shnatsel, line 130:
```suggestion
    /// Note that this default implementation will **not** be vectorized by the compiler automatically.
    /// For maximum performance you will need to override this function and implement it via
    /// an explicit batched conversion such as [`convert_to_f32_slice`](https://docs.rs/half/2.3.1/half/slice/trait.HalfFloatSliceExt.html#tymethod.convert_to_f32_slice)
```

src/block/samples.rs
```diff
@@ -122,31 +124,80 @@ pub trait FromNativeSample: Sized + Copy + Default + 'static {
...
+    fn from_f32s(from: &[f32], to: &mut [Self]) {
+        assert_eq!(from.len(), to.len(), "slices must have the same length");
+        for (from, to) in from.iter().zip(to.iter_mut()) {
+            *to = Self::from_f32(*from);
+        }
+    }
+
+    /// Convert all values from the slice into this type.
+    /// This function exists to allow the compiler to perform a vectorization optimization.
+    /// Note that this default implementation will **not** be vectorized by the compiler automatically.
```
Shnatsel, line 152:
```suggestion
    /// Note that this default implementation will be vectorized by the compiler automatically,
    /// provided that the CPU supports the necessary conversion instructions.
    /// For example, x86_64 lacks the instructions to convert `u32` to floats,
    /// so this will inevitably be slow on x86_64.
```

johannesvollmer:
great news, now we finally see the expected speedups in stable without setting a flag :) nice

```
cargo bench pixel format conversions
new read_image_rgba_f32_to_f16 ... bench:  18,413,400 ns/iter (+/- 2,459,747)
old read_image_rgba_f32_to_f16 ... bench:  29,149,950 ns/iter (+/- 3,587,253)
```

johannesvollmer:
Thanks for the review :) 

I forced `half` to the newest version, to avoid regressions, and added a unit test. 
Ready to merge. Any objections?

Shnatsel:
I would require `half` 2.2 but put `2.3.1` in the `Cargo.lock` to avoid breaking existing users by enforcing a higher MSRV, and add a note to the README about performance.

IMO a slower `f16` conversion is not as critical as the code failing to build in the first place.

Shnatsel:
> no compression used

I believe I changed those tests to read to `f32` rather than `f16`, which is why you're not seeing a speedup there. If you change them back to `f16` you will see a 9,000,000 ns/iter difference there too, because the format conversion is single-threaded.

Shnatsel:
I don't have a strong opinion on bumping `half`, however. 

I have no objections to merging this PR.

johannesvollmer:
okay
okay :)

one nice thing about the benchmarks: reading an image with f16 to f32 conversion is now as fast as reading f32 to f32 (no conversion). awesome!

johannesvollmer:
Good point about the MSRV. Didn't know that it changed. This repo has a CI test that checks for the MSRV still being valid - it seems this doesn't work correctly

johannesvollmer:
Added more benchmarks, everything looks as expected still. neat!

```
test read_f16_as_f16_uncompressed_1thread ... bench:  11,665,580 ns/iter (+/- 2,527,286)
test read_f16_as_u32_uncompressed_1thread ... bench:  11,732,710 ns/iter (+/- 957,454)
test read_f16_as_f32_uncompressed_1thread ... bench:  11,661,750 ns/iter (+/- 716,012)
test read_f16_as_f16_zip_nthreads         ... bench:  13,345,020 ns/iter (+/- 1,558,281)
test read_f16_as_f32_zip_nthreads         ... bench:  12,881,160 ns/iter (+/- 4,175,510)
test read_f16_as_f16_zip_1thread          ... bench:  28,832,260 ns/iter (+/- 2,584,587)
test read_f16_as_f32_zip_1thread          ... bench:  26,279,960 ns/iter (+/- 2,138,992)

test read_f32_as_f32_uncompressed_1thread ... bench:  17,843,730 ns/iter (+/- 1,008,381)
test read_f32_as_u32_uncompressed_1thread ... bench:  17,952,880 ns/iter (+/- 2,185,665)
test read_f32_as_f16_uncompressed_1thread ... bench:  17,965,450 ns/iter (+/- 2,524,674)
test read_f32_as_f32_zips_nthreads        ... bench:  26,873,920 ns/iter (+/- 3,032,381)
test read_f32_as_f16_zips_nthreads        ... bench:  26,641,840 ns/iter (+/- 2,400,515)
test read_f32_as_f32_zips_1thread         ... bench: 101,547,150 ns/iter (+/- 6,313,799)
test read_f32_as_f16_zips_1thread         ... bench: 100,998,820 ns/iter (+/- 6,737,638)
```

src/image/read/specific_channels.rs
```diff
@@ -279,30 +280,118 @@ pub struct OptionalSampleReader<DefaultSample> {
...
+
+    let total_sample_count = out_samples.len();
+    let batch_count = total_sample_count / batch_size;
+    let remaining_samples_count = total_sample_count % batch_size;
+
+    let len_error_msg = "sample count was miscalculated";
+    let byte_error_msg = "error when reading from in-memory slice";
+
+    // write samples from a given slice to the output iterator. should be inlined.
+    let output_n_samples = &mut move |samples: &[To]| {
```
Shnatsel, line 336: You can actually annotate closures with `#[inline]` if you want to hint to the optimizer that it should inline it.

src/block/samples.rs
```diff
@@ -122,31 +124,85 @@ pub trait FromNativeSample: Sized + Copy + Default + 'static {
...
+        for (from, to) in from.iter().zip(to.iter_mut()) {
+            *to = Self::from_f16(*from);
+        }
+    }
+
+    /// Convert all values from the slice into this type.
+    /// This function exists to allow the compiler to perform a vectorization optimization.
+    /// Note that this default implementation will be vectorized by the compiler automatically.
+    #[inline]
+    fn from_f32s(from: &[f32], to: &mut [Self]) {
```
Shnatsel, line 145: Is this `#[inline]` necessary? That would make a difference if the function operated on a single element, but since it processes a whole slice at once, I doubt the function call overhead will be at all noticeable.

Ditto for `fn from_f16s` and `fn from_u32s`.

Shnatsel: Ah nevermind. You're calling this only on slices of length 16 instead of the entire file because the API actually writes to an iterator, so this `#[inline]` is actually necessary.

src/image/read/specific_channels.rs
johannesvollmer: I like the idea, but when I tried it, it said that it's experimental to add attributes to expressions, so we'll have to wait

Shnatsel: I see. Looks good to go then!

johannesvollmer:
(sorry for not merging yet, I'm abusing this branch to fix the github workflow, it should have catched the MSRV breaking change)

Shnatsel:
As it stands, `Cargo.lock` does require Rust 1.70 but `Cargo.toml` does not, meaning that downstream users are free to configure the library to use older `half` with an older MSRV if they need to. I think that's a fair balance. It would be nice to call it out in the README.

johannesvollmer:
I'm not sure what you mean. If we allow `half = "2.3"`,  we should also raise our own `rust-version = 1.70.0` in the Cargo.toml, right? Do you mean we do that, and also hint at a workaround? The workaround being our users specify an older version of half and can then compile using rust 
 1.59? This makes sense, I think.

Shnatsel:
You can put `half = 2.2` into `Cargo.toml`, so when someone adds `exrs` as a dependency it will default to the latest at the time of the installation (currently 2.3.1) but will also allow downgrading to 2.2 if this is needed by the users for MSRV reasons.

And just don't put `rust-version` in there I guess, so you could have a `Cargo.lock` for development with the latest `half` for best performance, and also if anyone wants to run the benchmarks on the repo.

johannesvollmer:
The `Cargo.lock` is no longer in the repo, as is suggested for Rust libraries. But anyways, the plan still sounds good. I'll find out what the MSRV in the `Cargo.toml` actually means, and decide whether to put the Rust version into the `Cargo.toml`. Thanks for your help with this PR :)

johannesvollmer:
Actually, let's merge all of this except for the version upgrade of half. then release a major version with the minor performance improvements. then release 2.0.0 with the new version of half, including the new intrinsics

johannesvollmer:
`cargo-msrv verify` succeeds on my local machine... seems to be broken still

Shnatsel:
I am convinced that bumping semver for MSRV reasons alone is a bad idea, because now several crates using `exr` have to all manually upgrade in tandem. If one uses `exr` 1.x and the other uses 2.x they can no longer interoperate; and anyone can upgrade to 2.x only when *everyone* has upgraded to 2.x, splitting the ecosystem. Please don't release 2.x just because of an MSRV change.

johannesvollmer:
Hmmm maybe I shouldn't make spontaneous decisions at 3AM :D 

Yes, I see your point. On the other hand, most people will have specified the dependency to `exr = "1.5.6"` or similar in their project. So running `cargo update` will bump that to anything `<2.0.0` without asking, and then break their build if they do not use the absolutely newest Rust version. Is this expected behaviour? I thought it would be nice to explicitly opt-in to that, with more than just `cargo update`. However, I agree that splitting the ecosystem would not be good either.

johannesvollmer:
I've opened #217 and would like to continue the discussion there, as it seems more appropriate. I yanked version `2.0.0` for now, which I released earlier at 3AM before going to bed, so I have might not put enough thought into it back then, whoopsie

PR #193, opened 2023-01-20
Change `convert_current_to_little_endian` to return `Cow<[u8]>`

expenses:
This was a pretty easy change to make. I haven't profiled it but according to heaptrack, this reduces the number of allocations in a benchmark by half:

20230120_19h30m26s_grim
20230120_20h50m04s_grim

### DIFF
--- a/src/compression/mod.rs
+++ b/src/compression/mod.rs
@@ -15,6 +15,7 @@ mod b44;
 use crate::meta::attribute::{IntegerBounds, SampleType, ChannelList};
 use crate::error::{Result, Error, usize_to_i32};
 use crate::meta::header::Header;
+use std::borrow::Cow;
 
 
 /// A byte vector.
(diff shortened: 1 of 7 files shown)


src/compression/mod.rs
```diff
@@ -297,7 +303,7 @@ impl Compression {
 // FIXME this should really be done inside each compression method
 
 #[allow(unused)]
-fn convert_current_to_little_endian(bytes: Bytes<'_>, channels: &ChannelList, rectangle: IntegerBounds) -> ByteVec { // TODO is this really not already somewhere else?
+fn convert_current_to_little_endian<'a>(bytes: Bytes<'a>, channels: &ChannelList, rectangle: IntegerBounds) -> Cow<'a, [u8]> { // TODO is this really not already somewhere else?
```
johannesvollmer, line 306: While `Cow` is probably good performance-wise, I'm wondering if the code would be simpler in total if we were to change the signature to `fn convert_current_to_little_endian(bytes: &mut [u8], channels: &ChannelList, rectangle: IntegerBounds) { ... }`?

johannesvollmer:
Thanks for the suggestion :) a different PR incorporated these ideas

PR #194, opened 2023-01-20
Fully Parallel Block Reader/Decompressor/Interleaver

johannesvollmer:
sketch some ideas


johannesvollmer:
current approach is to spawn ALL the threads at once. the threads open the files individually. use the threadpool size to control how many file handles are opened simultaneously. however, this requires the closure which opens the file to be `Sync`, might be a problem

johannesvollmer:
@Shnatsel does this look like a good approach to you?

src/block/reader.rs
```diff
@@ -382,6 +382,55 @@ impl<R: ChunksReader> SequentialBlockDecompressor<R> {
     }
 }
 
+// note: this is just pseudocode, it will be split up into a lot of separate pieces to allow for customization
+fn read_all_blocks_fully_parallel<Read, PixelBlock>(
+    mut create_reader_at_byte: impl Sync + FnMut(u64) -> Result<Read>,
```
johannesvollmer, line 387: with the current approach, the closure must be immutable, because it needs to be sync

src/block/reader.rs
```diff
@@ -382,6 +382,55 @@ impl<R: ChunksReader> SequentialBlockDecompressor<R> {
...
+
+    for &chunk_index in offset_tables.iter().flatten() {
+        let meta = meta.clone();
+
+        thread_pool.spawn(move ||{
+            let meta = meta.as_ref();
+
+            let result = {
+                // TODO might want to do this on main thread such that the closure does not have to be sync?
+                let file_handle = create_reader_at_byte(chunk_index)?;
```
Shnatsel, line 420: I'm pretty sure you cannot read from the same file descriptor from multiple threads at once, and repeated calls to `open()` might be expensive.

What I'd do is have each thread have its own `file_handle`, all pointing to the same underlying file.

I wonder what happens if you use `thread_local` - will it complain that the data from it cannot be used on other threads, or will it create the `thread_local` on all threads and let you initialize it from the closure?

src/block/reader.rs
Shnatsel: I don't think requiring the closure to be `Sync` is going to be an issue.

src/block/reader.rs
```diff
@@ -382,6 +382,55 @@ impl<R: ChunksReader> SequentialBlockDecompressor<R> {
...
+        )?;
+    }
+
+    std::mem::drop(read_buffered); // TODO re-use when single threaded?
+
+    let meta = Arc::new(meta);
+    let thread_pool = threadpool::Builder::new().build();
+    let (sender, receiver) = ::flume::unbounded(); // TODO bounded? bounded to threadpool size?
+
+    for &chunk_index in offset_tables.iter().flatten() {
```
johannesvollmer, line 412: the tables should be sorted first to improve linear access

src/block/reader.rs
johannesvollmer: `create_reader` was intended to call `open` each time, inside each thread. this way, each thread has its own `file_handle`. alternatively, the closure can also open the file once and then call `try_clone` on the file handle. the closure has the responsibility that each call opens the same file at the specified location, but everything else can be changed. 

using `thread_local` would be something the closure could also do, I assume.

johannesvollmer: update: `try_clone` will not have it's own seek position :(

[docs](https://doc.rust-lang.org/std/fs/struct.File.html#method.try_clone)

johannesvollmer: for the initial implementation, I would prefer to open multiple file handles, and then benchmark

Shnatsel: Fair enough :+1:

johannesvollmer:
made the code compile somehow. required a lot of changes that i did not see coming.
here's the benchmark with the current code (opening a new file handle per block, which means hundreds per thread)

```
test read_single_image_uncompressed_rgba                ... bench:  16,190,040 ns/iter (+/- 1,368,617)
test read_single_image_uncompressed_rgba_fully_parallel ... bench:  41,465,800 ns/iter (+/- 8,681,578)
```

src/block/reader.rs
```diff
@@ -382,6 +384,154 @@ impl<R: ChunksReader> SequentialBlockDecompressor<R> {
...
+    //println!("starting thread pool work...");
+
+    // must be scoped because we want the multi byte source to contain byte slices, which have a lifetime
+    let result = thread_pool.scoped(|scope|{
+        for &chunk_location in sorted_chunk_offsets {
+            //println!("adding job for chunk at byte {}...", chunk_location);
+
+            let meta = meta.clone();
+            let sync_byte_source = sync_byte_source.clone();
+            let sender = sender.clone();
```
johannesvollmer, line 484: we might get away without any `Arc`s because we use a scoped thread pool, which means we can borrow for the whole scope

Shnatsel:
Benchmarking reading to `f16` instead of to `f32` might be more interesting. That way we'll see the effects of the format conversion being parallelized.

If you send me the file or include the file in the repo I can profile the code to see why it's slow. Or you can profile it yourself as I described in [the bounds checks article](https://shnatsel.medium.com/how-to-avoid-bounds-checks-in-rust-without-unsafe-f65e618b4c1e).

johannesvollmer:
the file is not special, I just needed a quick way to read from SSD, so I copied it to another Drive (the repository happened to be on my HDD)

currently no conversion is done at all (and adding it will require some more code, which is why it's not there yet)

johannesvollmer:
great to see you article online! nice read, and quite some surprising insights :o

johannesvollmer:
Got some better performance by avoiding to clone the file handle, using [`map_init` from rayon](https://docs.rs/rayon/latest/rayon/iter/trait.ParallelIterator.html#method.map_init) see this pseudo code:

```rust
chunk_offsets.into_par_iter()
        .map_init(
            || open_new_file_handle(),
            |file_handle, chunk_offset| {
                file_handle.seek(SeekFrom::Start(chunk_offset))?;
                process_chunk(meta.as_ref(), file_handle, pedantic, pixels_from_block)
            }
        )
        .try_for_each(|| sender.send(block))?;
```

johannesvollmer:
by the way, I'm pretty sure we can do parallel sample conversion with the current system right now. even if we can't find out how to implement the multiple file handles yet

src/block/reader.rs
johannesvollmer: we could also try using `try_clone`, which has no separate seek position, but then use `file.read_exact_at(offset)` which does not modify the seek position in the file handle. could it be faster?

see https://docs.rs/read_write_at/0.1.0/read_write_at/trait.ReadAt.html#method.read_exact_at

Shnatsel: Perhaps! I've never used that API so you'll just have to try it.

johannesvollmer:
with f16 conversion speed that high, and no major speedup by using multiple file handles, closing this for now

PR #213, opened 2023-07-01
Get <1.70 Working Again

michaelciraci:
Currently, `exrs` (and as a result, [image](https://crates.io/crates/image)) does not compile as a dependency for rust `<1.70` due to a breaking change with [half](https://github.com/starkat99/half-rs/issues/97):

```
cargo +1.69 test   
    Updating crates.io index
error: package `half v2.3.1` cannot be built because it requires rustc 1.70 or newer, while the currently active rustc version is 1.69.0
Either upgrade to rustc 1.70 or newer, or use
cargo update -p half@2.3.1 --precise ver
where `ver` is the latest version of `half` supporting rustc 1.69.0
```

The breaking change in [half](https://crates.io/crates/half) is documented [here](https://github.com/starkat99/half-rs/issues/97)

This merge requests bounds half to allow compilation again with rust `<1.70`.

In addition, `Cargo.lock` should be added to `.gitignore` since this is a package and not library: [https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries)


johannesvollmer (approved):
Hi! Thanks for taking the time :) I'll prepare the next release

johannesvollmer:
done :)
https://github.com/johannesvollmer/exrs/releases/tag/v1.6.5

PR #219, opened 2023-08-26
Don't use flume's default features

fintelia:
Notably, this eliminates a dependency on `syn` which improves compile time.


johannesvollmer:
Thank you for investigating and taking the time :) I'll prepare the release soon :)

johannesvollmer:
thanks for the heads-up! :)

PR #220, opened 2023-08-30
bump flume version for spin-rs dependency update

okey:
flume depends on spin-rs, and spin-rs prior to `v0.9.8` has a [concurrency bug](https://github.com/advisories/GHSA-2qv5-7mw5-j3cg)

flume updated its spin dependency in https://github.com/zesterer/flume/commit/9f67d0d2b4c57844f3b5d6ec26e6d9f4446dae93 and released that as `v0.11.0` in https://github.com/zesterer/flume/commit/ee6fe6adb98856edbda47d1cc867cbc255ff0d88

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ bit_field = "^0.10.1"          # exr file version bit flags
 miniz_oxide = "^0.7.1"         # zip compression for pxr24
 smallvec = "^1.7.0"            # make cache-friendly allocations        TODO profile if smallvec is really an improvement!
 rayon-core = "^1.11.0"         # threading for parallel compression     TODO make this an optional feature?
-flume = { version = "^0.10.9", default-features = false }              # crossbeam, but less unsafe code        TODO make this an optional feature?
+flume = { version = "^0.11.0", default-features = false }              # crossbeam, but less unsafe code        TODO make this an optional feature?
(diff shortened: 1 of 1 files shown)


johannesvollmer:
Thanks for taking the time to contribute :) Looks good, I'll merge that and publish a new version

okey:
@johannesvollmer Regarding the rust version check failure:

Locally, `rm -f Cargo.lock && cargo-msrv verify` and `rm -f Cargo.lock && cargo +1.59.0-x86_64-unknown-linux-gnu update && cargo-msrv verify` succeed but `cargo update && cargo-msrv verify` fails, probably because `image 0.24.4`  requires rust `1.61.0`

This PR could be updated to raise the MSRV of exrs, or by removing `cargo update` from the GitHub workflow, or by pinning the `image` more strictly. What do you think?

johannesvollmer:
The MSRV check failed a few times now, not only on this PR. I wan't able to figure out what exactly caused the failure. Seeing that the `image` crate uses a newer version than we do, we should raise our MSRV to `1.61.0`. If you want to contribute that, that would be awesome :)

okey:
@johannesvollmer I have updated this PR to raise the `exrs` MSRV to `1.61.0`

johannesvollmer:
Thanks for your time and efforts :)

PR #225, opened 2023-09-21
Updated dependencies

nadenf:


johannesvollmer:
Hi! Thanks for taking the time to contribute. However, it's better for users if we keep the versions at the lowest possible - it allows cargo to merge the dependencies with other dependencies in their project. Cargo will automatically use the newest version of each crate, because we only specify the minimum version using `^`. 

The reason that `half` is limited even more is to stay compatible with the older rust compilers

PR #227, opened 2023-10-07
sketch: read blocks on demand

johannesvollmer:


src/meta/attribute.rs
```diff
@@ -899,10 +899,19 @@ impl IntegerBounds {
...
+        subset.position.x() >= self.position.x()
+            && subset.position.y() >= self.position.y()
+            && subset.end().x() <= self.end().x()
+            && subset.end().y() <= self.end().y()
+    }
+
+    /// Returns whether the specified rectangle touches this rectangle.
+    pub fn intersects(self, other: Self) -> bool {
+        // https://stackoverflow.com/questions/2752349/fast-rectangle-to-rectangle-intersection
+        other.position.x() <= self.end().x()
```
remifontan, line 911: Considering the end is not contained in the bound, should it be `<` instead of `<=`?
I noticed that when loading single pixel at the border of a tile, 2 tiles get loaded. switching to `<` seems to give the expected behaviour.

remifontan: I suspect it may be the same for `fn contains` , considering the bound is "exclusive", `end()` is not included in the bound, is that right?

remifontan: I would expect this test to pass:
```rust
let a = IntegerBounds::new( (0, 0), (10, 10));
let b = IntegerBounds::new( (10, 0), (1, 1));
assert!(!a.intersects(b));
```

johannesvollmer: that would make for a perfect first test :)

johannesvollmer:
idea: make parallel block decompressor usable independently of the other decoders

PR #231, opened 2024-01-30
Allow `half 2.3` (and higher) to be used while also testing a 1.61 MSRV

### LINKED ISSUE #230: Strange `half <2.3.0` upper-bound breaks compatibility with other crates
Consider this sample crate, which compiles:

```toml
[package]
name = "broken"
version = "0.1.0"
edition = "2021"

[dependencies]
exr = "1.6.4"
rerun = "0.12"
```

The `half <2.3.0` bound introduced https://github.com/johannesvollmer/exrs/pull/213 (to work around https://github.com/starkat99/half-rs/issues/97) prevents this crate to be used with crates - [e.g. `re_log_types` from `rerun`](https://crates.io/crates/re_log_types/0.12.1/dependencies) - which depend on `half ^2.3.1`. If the `Cargo.toml` sample above is bumped to `exr 1.6.5`, those bounds are now incompatible according to `cargo`:

```
error: failed to select a version for `half`.
    ... required by package `exr v1.6.5`
    ... which satisfies dependency `exr = "^1.6.5"` of package `broken v0.1.0 (broken)`
versions that meet the requirements `>=2.1.0, <2.3` are: 2.2.1, 2.2.0, 2.1.0

all possible versions conflict with previously selected packages.

  previously selected package `half v2.3.1`
    ... which satisfies dependency `half = "^2.3.1"` of package `re_log_types v0.12.0`
    ... which satisfies dependency `re_log_types = "^0.12.0"` of package `rerun v0.12.0`
    ... which satisfies dependency `rerun = "^0.12"` of package `broken v0.1.0 (broken)`

failed to select a version for `half` which could resolve this conflict
```

I don't see why `exr` needs to take a stake in `half`'s MSRV bump. If you wish to maintain a low MSRV based on minimal versions, I recommend leaving the semver-compatible upper bound open (`^2.1.0`) and testing your MSRV with a fixed `cargo update -p half --precise 2.1.0` or `cargo +nightly -Zminimal-version generate-lockfile` in the CI.

It looks like that was intended to be done in https://github.com/johannesvollmer/exrs/commit/42a92aa4cabde3cb0562712a251d21f161171a86 but that ended up reverted in https://github.com/johannesvollmer/exrs/commit/3d43e30a678b5c89a229d4e71ea4fca636378a17 without commit message, so I can only assume that at least `rust-version = "1.70.0"` broke building on Rust <1.69 even with the intended/documented workaround.

MarijnS95:
Fixes #230

Instead of outright failing to compile when the `exr` crate is compiled together with a crate that selects a `half` crate version outside of the `">=2.1.0, <2.3"` range, allow anything `">=2.1.0"` but suggest people how they can _opt in_ (unfortunately it's not an opt-out) of the original `1.61` MSRV.

The difference with https://github.com/johannesvollmer/exrs/commit/42a92aa4cabde3cb0562712a251d21f161171a86 is that this does not bump `rust-version = "1.70"` which would break the suggested `half = "2.1.0"` workaround. This workaround doesn't work anyway because `"2.1.0"` allows anything `">=2.1.0, <3.0.0"`. Instead the suggestion should use the version bound provided here, or an explicit `"=2.1.0"` pinned version.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -37,8 +37,11 @@ jobs:
         run: cargo test --verbose
 
   verify-msrv:
-    runs-on: ubuntu-latest
-    name: verify minimum supported rust version (ubuntu)
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+    runs-on: ${{ matrix.os }}
+    name: verify minimum supported rust version without cache (takes longer, but caching produces unexpected behaviour)
     timeout-minutes: 30
 
     # we are using the `cargo-msrv` app
@@ -47,14 +50,15 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Install foresterre/cargo-msrv without cache (takes longer, but caching produces unexpected behaviour)
+      - name: Install foresterre/cargo-msrv
         run: cargo install cargo-msrv
 
-      - name: Verify the Rustc version declared in `cargo.toml` without cache (takes longer, but caching produces unexpected behaviour)
-        run: |
-          rm -f Cargo.lock
-          cargo update
-          cargo-msrv verify
+      - uses: dtolnay/rust-toolchain@nightly
+      - name: Generate Cargo.lock with minimal-version dependencies
+        run: cargo -Zminimal-versions generate-lockfile
+
+      - name: Verify the Rustc version declared in `Cargo.toml` with `minimal-versions`
+        run: cargo-msrv verify
 
   # github actions does not support big endian systems directly, but it does support QEMU.
   # so we use cross-rs, which uses QEMU internally.
@@ -94,16 +98,16 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    
+
     - name: Cache Cargo Dependencies
       uses: Swatinem/rust-cache@v1.3.0
-      
+
     - name: Add wasm32 Target
       run: rustup target add wasm32-unknown-unknown
-      
+
     - name: Build without default features
       run: cargo build --verbose --no-default-features --target wasm32-unknown-unknown
-      
+
     - name: Run tests without default features
       run: cargo test --verbose --no-default-features
 
--- a/README.md
+++ b/README.md
@@ -28,11 +28,6 @@ Features include:
 
 ### Current Status
 
-> [!TIP]
-> MSRV Note: This crate is now on Rust `1.70.0`. If you want to use the
-> newest version of `exrs` with an older Rust version, you can still do that:
-> By specifying a version `half = "2.1.0"` in your project, the crate will work with Rust `1.59.0`.
-
 This library has matured quite a bit, but should still be considered incomplete.
 For example, deep data and DWA compression algorithms are not supported yet.
 
@@ -162,6 +157,9 @@ __What we can do:__
 
 
 ### Usage
+
+> [!TIP]
+> If you want to use the newest version of `exrs` with an older Rust version, you can still do that, by forcing Rust to use a an older version of the `half` crate via `cargo update -p half --precise 2.2.1`, or downgrade all dependencies via `cargo +nightly -Zminimal-versions generate-lockfile`.  `half 2.3.0` and higher have an MSRV above 1.61.
 
 Add this to your `Cargo.toml`:
 ```toml
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-[![Rust Docs](https://docs.rs/exr/badge.svg)](https://docs.rs/exr) 
+[![Rust Docs](https://docs.rs/exr/badge.svg)](https://docs.rs/exr)
 [![Rust Crate](https://img.shields.io/crates/v/exr.svg)](https://crates.io/crates/exr)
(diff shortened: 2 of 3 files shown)


MarijnS95:
Looking at the CI failure, `half` is properly downgraded to `2.1.0` (we might run `2.2.x` too?) but some other dependency also breaks MSRV (the check succeeds with `-Zminimal-versions`!). `cargo-msrv`, in all its wisdom fails to report what crate is at fault (i.e. the error returned by `cargo check`).

Do we want to look into what dep this is, or remove the mentions of downgrading `half` and only suggest `-Zminimal-versions`?

johannesvollmer:
Given that our CI guarantees that 1.61 indeed compiles with `-Zminimal-versions`, simply using that sounds reasonable to me. We will be notified if any dependency breaks the build, correct?

I'd like to make sure that the `image`crate still compiles after this change. We should investigate whether they can still compile with the workaround, and where it would fail without the workaround.

MarijnS95:
> We will be notified if any dependency breaks the build, correct?

Yeah, though none of those can change if minimum version requirements in `exr`'s `Cargo.toml` don't change (unless versions are yanked and `cargo` picks newer ones "automatically").

> I'd like to make sure that the `image`crate still compiles after this change. We should investigate whether they can still compile with the workaround, and where it would fail without the workaround.

Do you expect me as submitter to test that? I can take a look.

README.md
```diff
@@ -28,13 +28,17 @@ Features include:
 
 ### Current Status
 
-This library has matured quite a bit, but should still be considered incomplete. 
+> [!TIP]
+> MSRV Note: This crate is now on Rust `1.70.0`.
```
MarijnS95, line 32: This wording might be somewhat misleading, as there is currently no CI validating for `1.70`.

README.md
```diff
@@ -28,13 +28,17 @@ Features include:
...
+> If you want to use the newest version of `exrs` with an older Rust version, you can still do that, by forcing Rust to use a an older version of the `half` crate via `cargo update -p half --precise 2.2.1`.  `half 2.3.0` and higher have an MSRV above 1.61.
```
MarijnS95, line 33: The suggested workaround is different from what is being tested in the CI, but it's a tad silly because `-Zminimal-versions` requires _nightly Rust_ (on the topic of supporting older MSRV)...

That only really works if users are using `Cargo.lock` and okay with using a _nightly_ toolchain to generate that (maybe on a different machine or CI if they're locked to a specific version at build-time only). `cargo update --precise ...` is a bit more approachable, and could be used in the `exrs` CI instead?

MarijnS95:
Okay I'm spinning up some `image` CI runs at https://github.com/MarijnS95/image/actions with this change. They'll likely fail and I'll end up requesting a clarification to generating their `Cargo.lock.msrv` because it seems to have been a snapshot at some point in time - possibly with manual `cargo update --precise ...` - as it is neither identical to one generated with `-Zminimal-versions`.

Perhaps they should employ the same `-Zminimal-versions` check in their CI rather than some open-coded lock file that's hard to contribute to.

MarijnS95:
Opened in https://github.com/image-rs/image/issues/2120.

README.md
johannesvollmer: I would be fine with just suggesting the existence of `-Zminimal-versions`, even if using that temporarily requires a nightly cargo version. I would assume that people struggling with an older compiler know these special approaches.

As you seem to know about this topic, I'd trust your input on that.

On a different note, I would ask you to put this whole section into the readme section where the installation command is shown. Ok?

johannesvollmer:
> Do you expect me as submitter to test that? I can take a look.

I didn't expect it, but I'm amazed that you've gone ahead already. Thanks for taking the time!

.github/workflows/rust.yml
```diff
@@ -47,14 +50,15 @@ jobs:
...
-        run: |
-          rm -f Cargo.lock
-          cargo update
-          cargo-msrv verify
+      - uses: dtolnay/rust-toolchain@nightly
+      - name: Generate Cargo.lock with minimal-version dependencies
+        run: cargo -Zminimal-versions generate-lockfile
+
+      - name: Verify the Rustc version declared in `Cargo.toml` with `minimal-versions`
+        run: cargo-msrv verify
```
johannesvollmer, line 61: awesome

README.md
johannesvollmer: Yes, I'd consider just removing the first sentence. Especially since it was always possible to use this crate with the newest version of Rust.

README.md
```diff
@@ -163,13 +167,13 @@ Add this to your `Cargo.toml`:
...
 # and better runtime performance
 [profile.release]
 lto = true
 ```
 
-The master branch of this repository always matches the `crates.io` version, 
+The master branch of this repository always matches the `crates.io` version,
 so you could also link the github repository master branch.
 
 ### Example
```
johannesvollmer, line 179: please put [TIP] section here :)

johannesvollmer (commented):
thanks again for your contribution, great work!

README.md
MarijnS95: Let's showcase both commands in the readme, it's a `[!TIP]` block anyway.

Moved under `### Usage`!

johannesvollmer (approved):
Excellent.

johannesvollmer:
I'll go on to prepare a new release.

johannesvollmer:
done, see https://crates.io/crates/exr/1.72.0

PR #241, opened 2024-10-11
add rayon feature

oligamiq:


Cargo.toml
```diff
@@ -44,23 +44,43 @@ walkdir = "2.3.2"         # automatically test things for all files in a directo
...
+required-features = ["rayon"]
 
 [[bench]]
 name = "pixel_format_conversion"
 harness = false
+required-features = ["rayon"]
 
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
```
fintelia, line 83: This isn't needed anymore, since rustdoc [recently added it as the default](https://github.com/rust-lang/docs.rs/pull/2390/files).

fintelia:
Not a maintainer, but FYI this is an API breaking change because any downstream code relying on the now cfg'd out methods will no longer compile. If rayon was left as a default feature, this change could be done in a backwards compatible way

oligamiq:
All the tests have passed, but indeed, this might be a breaking change.

johannesvollmer:
Thanks for taking the time to contribute! The changed you proposed seem solid. I would have solved it differently, but I don't have the time to do that:

As the calls to rayon are quite few and isolated, it should be possible to keep the API as-is. Then at runtime falling back to sequential processing if the rayon feature is disabled. This would require some more complicated code changed, but it would avoid us having to sprinkle `cfg` attributes in multiple unrelated code sections, which seems hard to maintain in my eyes (and is not backwards compatible).

The backwards incompatible changes would require major version upgrade, `exr 2.0`, which makes it harder for people to use the newest versions, as they have to manually update the version in their projects.

What do you think?

oligamiq:
I think it’s possible to maintain all the public structs and APIs while marking them as deprecated and falling back. I'll be busy for a few days, but after that, I can give it a try if you'd like.

johannesvollmer:
Checked the current code and found something problematic, there is at least one public function that exposes a rayon item at compile time: `pub fn new_with_thread_pool`. Someone might rely on this function, it references the rayon thread pool in the function type signature. This means we cannot get around a breaking change. Please correct me if I'm wrong.

oligamiq:
That's right. I share the same opinion. While it's possible to replace the dependent types with generics, it would cause some significant breaking changes. In the case of removing Rayon from image-rs, it's acceptable to have Rayon set to default ON. However, setting Rayon to default ON is generally not advisable.

oligamiq:
Since there seem to be a lot of `issues` left, how about removing rayon from the default feature list at the time of a major version update?

johannesvollmer:
after thinking about this, I'm happy to merge this. (of course we need to fix the 3 small conflicts, but that's not a big deal.) thanks for your contribution!

.github/workflows/rust.yml
```diff
@@ -20,6 +20,10 @@ jobs:
       run: cargo build --verbose
     - name: Run tests
       run: cargo test --verbose
+    - name: Build with rayon feature
+      run: cargo build --features rayon --verbose
+    - name: Run tests with rayon feature
+      run: cargo test --features rayon --verbose
```
johannesvollmer, line 26: we might need to invert this to run without default features since rayon will be enabled by default

Cargo.toml
```diff
@@ -32,7 +32,7 @@ half = "2.1.0"                 # 16 bit float pixel data type
 bit_field = "^0.10.1"          # exr file version bit flags
 miniz_oxide = "^0.7.1"         # zip compression for pxr24
 smallvec = "^1.7.0"            # make cache-friendly allocations        TODO profile if smallvec is really an improvement!
-rayon-core = "^1.11.0"         # threading for parallel compression     TODO make this an optional feature?
+rayon-core = { version = "^1.11.0", optional = true }         # threading for parallel compression
 flume = { version = "^0.11.0", default-features = false }              # crossbeam, but less unsafe code        TODO make this an optional feature?
```
johannesvollmer, line 36: flume has been removed, just remove this line to solve the conflict

johannesvollmer:
@fintelia what's the thread pool choice of the image crate? i'm assuming the rayon feature will be turned off in the image crate. that results in 50%-80% loss in performance for exrs. but I think you still have threading in the image crate?

maybe exrs can adapt to other parallelism crates than rayon? or maybe std has a thread pool by now and I didn't notice?

fintelia:
Yes. After this lands, `image` will switch to only enabling the rayon feature on `exr` when its own rayon feature is enabled

oligamiq:
I accepted the CI-related reviews and resolved the conflicts.

johannesvollmer (approved):
Thanks for taking the time for the follow up! I appreciate your contributions.

PR #250, opened 2025-09-06
use errors everywhere instead of panicking when casting primitive numbers

johannesvollmer:


johannesvollmer:
note: before merging, change base to main.

PR #262, opened 2025-11-19
DWAA/B Support

virtualritz:
# DWA Status

* The current DWAA/B decoder is functionally complete but not bit‑exact with the OpenEXR reference: the last scanline still differs by ±1 half‑float ULP in a few pixels (numerically that means values differ by less than 0.001 from reference).
   > My take is that this is not a blocker because if you use DWAA/B (lossy) compression you are ok with variance in this range for starters and if you are a small studio that can't pay TDs to keep an OpenEXR C/C++ deployment up to date and use Rust instead because `cargo != pain` you will gladly ignore this even more so.

* Zigzag unpacking, IDCT constants, CSC math, and the `float`→`half` conversion now mirror OpenEXR's scalar pipeline. Remaining differences stem from the surrounding `LossyDctDecoder` loop (block ordering, constant block shortcut, rounding order).
* SIMD paths (SSE2/AVX/NEON) from OpenEXR are not ported. All math runs on scalar `f32` and Rust’s `half` conversions, so performance and rounding differ from the C implementation.
* We do not implement OpenEXR's `convertFloatToHalf64_*` batching or its aligned row‑block staging. Our decode loop writes scanlines directly, which changes the reduction order and rounding.
* The codec still lacks an encoder; compression is a stub returning `Error::unsupported`.
* A DWAA/B encoder now exists and can roundtrip existing non‑lossy inputs, but it has not been validated against OpenEXR bitstreams beyond functional testing.
* Improvements needed for parity:
  - Port `LossyDctDecoder_execute` verbatim (including `DctCoderChannelData`, constant block detection, row block allocation, and planar writeback).
  - Reuse OpenEXR’s half conversion helpers (`float_to_half_int`) everywhere and consider calling the SIMD paths via intrinsics to match rounding and throughput.
  - Add conformance tests that compare against the OpenEXR decoder for multiple images and codecs `(DWAA`/`DWAB`, `F16`/`F32`) to guard against regressions.


johannesvollmer:
This was implemented in a different branch :) Thanks still!

virtualritz:
Very cool!

PR #270, opened 2026-03-09
Add prepack script to fix broken npm publish

akre54:
The `exrs` package (v1.0.2) fails to install due to an unresolvable `file:../pkg` dependency reference. The root cause is that `exrs-raw-wasm-bindgen` was listed in `bundledDependencies` but the full build was never run before publishing, so nothing was actually bundled in the tarball.

Adds a `prepack` lifecycle script that runs `npm run build` (wasm-pack → npm install → tsc) before packing, ensuring `node_modules/exrs-raw-wasm-bindgen` is populated and bundled correctly.

### DIFF
--- a/exrs-wasm/js/package.json
+++ b/exrs-wasm/js/package.json
@@ -34,6 +34,7 @@
   "scripts": {
     "prebuild": "cd .. && wasm-pack build --target web",
     "build": "npm i && tsc",
+    "prepack": "npm run build",
     "test": "npm run build && vitest run",
     "format": "prettier --write ."
   },


johannesvollmer (approved):
Code looks good. I will run it locally to test and then publish the package and then merge afterwards

johannesvollmer:
Published the new package. Unpacked size is now 595 kB - I think it worked:
image

akre54:
Testing now, thanks!

akre54:
Claude says:

> Still failing - v1.0.3 has the same `file:../pkg` reference. The package.json in the npm publish still contains:
> 
> ```json
> "exrs-raw-wasm-bindgen": "file:../pkg"
> ```
> 
> This needs to be removed or changed. The `pkg/` directory contents should be bundled directly into the package, not referenced as a separate local dependency.
> 
> **Quick fix:** In your exrs repo, check if `pkg/` is being included in the npm tarball. If so, change the dependency to:
> 
> ```json
> "exrs-raw-wasm-bindgen": "file:./pkg"
> ```
> 
> (note: `./pkg` not `../pkg`)
> 
> Or better, just inline the wasm-bindgen exports directly without a separate package reference.

johannesvollmer:
As my screenshot shows, the package is correctly bundled in the tarball in a node_modules folder. But we can't simply change the path in the package.json, because then local builds would not be possible anymore. 

You can try to fix the npm package setup of the library, or maybe claude knows how to make both local and published workflows work at the same time.

johannesvollmer:
If you find a well maintainable way to "inline" the wasm package, you can do that, but my previous attempts to achieve that were all cumbersome and fragile by patching the build output files. I would rather just publish the internal wasm package if there is no better way.

akre54:
Ah okay, I'm remembering we had a [thread](https://github.com/akre54/exrs/pull/1#issuecomment-3781115806) on this in #263. Apparently `bundledDependencies` works with npm but Yarn Berry (v3+) doesn’t fully support it. It tries to resolve `file:../pkg` relative to the consumer's project, not the exrs package.

We could either change the build script ...

```bash
# In build script:
wasm-pack build --target bundler --out-dir js/pkg
rm js/pkg/package.json  # Remove the nested package.json
```

... then import directly from `./pkg/` in the wrapper code instead of using a dependency.

Or alternatively, publish `exrs-raw-wasm-bindgen` as a separate npm package ...
 
```json
// exrs/package.json
{
  "dependencies": {
    "exrs-raw-wasm-bindgen": "^1.0.0"  // version, not file path
  }
}
```

... then use npm workspaces or yarn workspaces for local development in the exrs repo:

```json
// root package.json (monorepo style)
{
  "workspaces": ["exrs", "pkg"]
}
```

Issue #235, opened 2024-06-27
Implement a poor man's `exrheader`


stevenparkerco3:
## What can be improved or is missing?
To help expand on the example applications, it would be great to have an example application which can output EXR Header information in the same/similar format to OpenEXR's `exrheader` application. I believe this would help developers explore the exrs API's a bit further and lower the barrier to entry. In line with what OpenImageIO calls their "Poor man's exrheader" tool using `iinfo`, it would be nice to have a similar example in `exrs`.

## Implementation Approach
I'm still learning Rust, and pretty terrible at it, but have the following Gist which is semi-functional:
https://gist.github.com/stevenparkerco3/d49b21072670312443521931eb698e46

johannesvollmer:
Yes, we also thought this might be useful. We planned to publish some simple tools similar to this as real binary creates, such that everyone can install them. 

Unfortunately, no one had the time to actually do it. Feel free to proceed.

I have one tip for you, which is to checkout the more low-level exrs api (the exr library uses this to construct the more complex types that you use currently)

johannesvollmer:
Are there some specific improvements in the library that would make your task easier? Do you want to 100% precisely replicate the OpenEXR output, or do you want to make a better one?

Here are more thoughts on your current status:

> Header flags HEX Output

Is this actually useful for the user to see the hex codes? I think it would be much more useful to a human user to display the [fields of the version struct](https://github.com/johannesvollmer/exrs/blob/861f0cf1791e83297f2d15b5c4230380bb7f20c1/src/meta/mod.rs#L68-L88).

> Some core header values are not parsed. Right now, it's just a hard coded list of some values I thought were important, but should extend to ALL values

We might want to add to the library a visitor pattern that visits each attribute. Do you think that makes sense? Something like:

```rs
header.visit_all_fields(|name, value|{
    println!("{} (type {}): {}", ...);
})
```

johannesvollmer:
https://github.com/johannesvollmer/exrs/pull/236/files

here is a prototype of the function, if you think it's useful, we can merge that

PR #271, opened 2026-03-10
refactor: Restructure repo into a cargo workspace

vvzen:
Based on the notes at https://github.com/johannesvollmer/exrs/issues/264 and the small chat in https://github.com/johannesvollmer/exrs/issues/235.

~Marking as "draft" until I will have done a few more tests locally to be sure I didn't mess anything up.~

## Notes (as of 2026-03-10)

You can run `cargo check` and `cargo test` successfully from the workspace root and from each subpackage.

You can also run some examples, like this:
```bash
cargo run --example 7_write_raw_blocks && cargo run --example 8_read_raw_blocks
```

The only warning I see is the same warning that I see also in the current master branch (with HEAD at 2f635aa):
```
warning: /home/vv/dev/oss/exrs/lib/exrs/Cargo.toml: unused manifest key: lib.plugin
```

I haven't done more in-depth testing, let me know if you other specific things that you'd expect to be working fine! 
I can imagine that there might be more bits off once you get to publish the crate, I never went the process so I don't know how that would work. I guess you might at least need to pass a specific `--package` to the `publish` invocation (see https://doc.rust-lang.org/cargo/commands/cargo-publish.html#package-selection)


Cargo.toml
```diff
@@ -111,3 +36,10 @@ overflow-checks = false
 lto = true
 debug = true
 codegen-units = 1
+
+# NOTE: When using per-package profile some keys (like `lto`) are currently not supported inside workspaces, see:
```
johannesvollmer, line 40: `LTO` seems like a pretty important optimization for this crate specifically, as we use a lot of generics and closures. Is it now not active anymore?

johannesvollmer:
Thanks for this suggestion! Nice to see that the old commands still work in the project root.

Cargo.toml
johannesvollmer: Wait, why is this even here? Shouldn't this be in the wasm crate cargo toml instead?

johannesvollmer:
This PR only moves the source folder. Given that we want to add some command line tool as a third subcrate with its own tests, it would probably be cleaner if we also move the existing tests, examples, and benches into the `lib/exrs` subfolder. Otherwise, the tests of the command line tool will mix with the tests of the core library, which does not seem optimal to me. What do you think?

lib/exrs-wasm/src/lib.rs
johannesvollmer, line 1: This file was not properly formatted, because it was previously in a nested crate (and cargo does not work like that, so the formatter ignored the whole subcrate). Now it's recognized as part of the workspace which means the formatter is rejecting the file if we don't format it.

vvzen:
> [...]  it would probably be cleaner if we also move the existing tests, examples, and benches into the lib/exrs subfolder. Otherwise, the tests of the command line tool will mix with the tests of the core library, which does not seem optimal to me.

I think it makes senses for the `tests` and the `benches`!
Definitely you wouldn't want to have them mixed.

For the `examples`, right now having the `examples` dir at the root of the workspace allows you to run things like `cargo run --example 7_write_raw_blocks` and I don't know if that would be possible if the examples were in a subdir, since I have a feeling that `cargo` doesn't allow to configure where the examples live. But if that's an okay trade off, I think it's fine to move everything!

Cargo.toml
vvzen: If you look at the field, it's named `[profile.release.package.exrs-raw-wasm-bindgen]` so it's specified at the workspace level (cargo will tell you off if you don't!) but it targets the sub-package. As for `lto`, I need to triple check, but when I searched for it it seems that it's not supported when using workspaces and subpackages. I'll have another read at https://github.com/rust-lang/cargo/issues/8264

vvzen: So after re-reading the related issue, I have a sense that probably we can still specify the custom `[profile.release]` at the sub-package level within the workspace, and it should honor it if people are just pulling that sub-package.

However, if you build from _within_ the workspace, you'd get the `warning: profiles for the non root package will be ignored, specify profiles at the workspace root:` warning that I originally got.
The thing is that I don't have an easy way to check if this hypothesis is correct, I guess `lto` should change the binary size in a way that is easy to detect but I'd need to setup a project that pulls in just `exrs-raw-wasm-bindgen` as a dependency, I guess? Tips are welcome!

vvzen: Ok, so, just for the records, with this diff:
```diff
diff --git a/Cargo.toml b/Cargo.toml
index 073d313..2de4664 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,4 +42,4 @@ codegen-units = 1
 # - https://github.com/rust-lang/cargo/issues/8264#issuecomment-635424772
 [profile.release.package.exrs-raw-wasm-bindgen]
 opt-level = "s"
-# lto = true
+lto = true
```

I get this:
```
$ cargo build
error: failed to parse manifest at `/home/vv/dev/oss/exrs/Cargo.toml`

Caused by:
  `lto` may not be specified in a `package` profile
```

johannesvollmer:
It would be an acceptable tradeoff for me to leave the examples at top level, or alternatively it would also be okay for me to loose the ability to launch single examples easily (I assume most people only read them and rarely run them)

Cargo.toml
vvzen: Ok so I think the TL;DR of that cargo issue is this, which I originally overlooked:
> However, if subcrate is a binary crate published to crates.io, then the profile section will be used when someone cargo installs that crate. Moving the profile section to the workspace root would remove the settings when cargo installing the crate downloaded from crates.io.

So what you could get is that folks that install the wasm crate via `cargo install --git https://github.com/johannesvollmer/exrs --bin exrs-raw-wasm-bindgen` (or `cargo install exrs-raw-wasm-bindgen` ?) would get the `lto = true` optimization. But right now that crate is a **lib**, so I don't know if there's any workflow where `cargo install`ing a lib has any value.


Out of curiosity I checked if the `lto = true` would be honored if folks were to just adding your crate to their own. So I created an empty `bin` crate that just pulls your crate..

e.g.: `cargo init` on an empty dir, and then this `Cargo.toml` :
```toml
[package]
name = "exrs-wasm-test"
version = "0.1.0"
edition = "2024"

[dependencies]
exrs-raw-wasm-bindgen = { path = '/home/vv/dev/oss/exrs/lib/exrs-wasm' }
```

Then I tried specifying the `lto = true` at the sub-crate level only here in `exrs`:
```diff
diff --git a/Cargo.toml b/Cargo.toml
index 073d313..5da2f8f 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,10 +36,3 @@ overflow-checks = false
 lto = true
 debug = true
 codegen-units = 1
-
-# NOTE: When using per-package profile some keys (like `lto`) are currently not supported inside workspaces, see:
-# - https://github.com/rust-lang/cargo/issues/15262
-# - https://github.com/rust-lang/cargo/issues/8264#issuecomment-635424772
-[profile.release.package.exrs-raw-wasm-bindgen]
-opt-level = "s"
-# lto = true
diff --git a/lib/exrs-wasm/Cargo.toml b/lib/exrs-wasm/Cargo.toml
index 3dd5911..8cb4e08 100644
--- a/lib/exrs-wasm/Cargo.toml
+++ b/lib/exrs-wasm/Cargo.toml
@@ -16,3 +16,7 @@ half = "2.1"
 
 wasm-bindgen = "=0.2.100"
 console_error_panic_hook = "0.1"
+
+[profile.release]
+opt-level = "s"
+lto = true

```

And then, when compiling in the other crate I got these file sizes:
```bash
$ cd /tmp/exrs-wasm-test
$ cargo clean && cargo build --release
$ find target/release -maxdepth 2 -name 'libexrs_raw_wasm_bindgen*' -exec du -sh {} \;
1.2M    target/release/deps/libexrs_raw_wasm_bindgen.rlib
428K    target/release/deps/libexrs_raw_wasm_bindgen.so
```

Trying w/ the current code at b505917 (so only specifying it at the workspace level, and without `lto = true`), I get the same exact results:

```bash
$ cd /tmp/exrs-wasm-test
$ cargo clean && cargo build --release
$ find target/release -maxdepth 2 -name 'libexrs_raw_wasm_bindgen*' -exec du -sh {} \;
1.2M    target/release/deps/libexrs_raw_wasm_bindgen.rlib
428K    target/release/deps/libexrs_raw_wasm_bindgen.so
```

So it doesn't seem like it gets honored.

vvzen: For the final test, I also compared the before/after this PR when compiling from the workspace.
Before this PR we have:

```bash
$ cd /tmp/exrs/exrs/exrs-wasm
$ cargo build --release
$ find target/release -maxdepth 1 -name 'libexrs_raw_wasm_bindgen*' -exec du -sh {} \;
4.0K    target/release/libexrs_raw_wasm_bindgen.d
3.6M    target/release/libexrs_raw_wasm_bindgen.rlib
452K    target/release/libexrs_raw_wasm_bindgen.so
````

After this PR we have:
```bash
$ cd ~/dev/oss/exrs/lib/exrs-wasm
$ cargo clean && cargo build --release
$ cd ../..
$ find target/release -maxdepth 1 -name 'libexrs_raw_wasm_bindgen*' -exec du -sh {} \;
3.0M    target/release/libexrs_raw_wasm_bindgen.rlib
432K    target/release/libexrs_raw_wasm_bindgen.so
4.0K    target/release/libexrs_raw_wasm_bindgen.d
```

So, surprisingly, it seems that the file size actually got smaller, even if it just barely.
So the ultimate TL;DR might be: maybe `lto = true` is doing so little here that it could be okay to ignore it altogether ?

vvzen:
Quick one: regarding the failing MSRV check, I think that the new one (after 68a4a83) might be `1.84.1`, at least according to what `cargo msrv find -- cargo check -p exr` says. I guess that's because of the `resolver = "3"` in the workspace toml: https://doc.rust-lang.org/edition-guide/rust-2024/cargo-resolver.html . If you want to support older versions, we can look what `resolver = "2"` would bring.

vvzen:
> It would be an acceptable tradeoff for me to leave the examples at top level, or alternatively it would also be okay for me to loose the ability to launch single examples easily (I assume most people only read them and rarely run them)

In d377e54 and a064a24 I moved benchmarks and tests in the `lib/exrs` dir, let me know if you want to move the examples too or not!

vvzen:
No pressure, but when you have time I think this is ready for another round of CI and review!

Cargo.toml
johannesvollmer: I think including this lib in your crate will depend on your crate's flags. This LTO flag would only affect the benchmarks in the repo. We should compare with LTO versus without LTO when running the benchmarks. I can do it probably, but later.

vvzen: Sounds good. I can give it a try later too if I have a bit of time!

johannesvollmer:
I have started to update this pr branch from main, stay tuned

PR #272, opened 2026-06-30
DWA Decomp for viewing and saving in new file. (OpenEXR translation)

zinedinp:
Added DWAA and DWAB decompression to load files, no writing.

Documentation Link was down.


johannesvollmer:
Hello, thank you for the proposal. Do you need some support?

PR #273, opened 2026-07-01
Dwa decomp

zinedinp:
I apologise for the last commit. When coding late at night, you can get a bit careless and just want to see the ‘working’ patch straight away, even though it feels like nothing matches OpenEXR, just viewing somehow worked.

Your documentation here: https://github.com/johannesvollmer/exrs/pull/262
helped me. (I should have searched for it erarlier)

Changes since the last commit
Removed Pi inaccuracy
Corrected the formula for linear
Added F32/U32 handling
Added CSC handling
SMID unsafe Rust for the tryhards that want 1to1 bits

I’d view the SIMD-difference as an intentional ambiguity in OpenEXR to prioritise speed, as these are different calculations.

If you see anything let me know, and thanks for the help you had offered.


tests/images/valid/custom/dwa_csc/generate.py
johannesvollmer, line 1: This would add another dependency (py) to the project. it's not a huge issue, since the generated files are added to the repo. but since there are rust bindings to the C++ OpenEXR library, it would be possible to do this in rust (I think we might even have them as devdependencies already).

 mainly, it would make it easier to generate more special cases in the future. this also might allow us to avoid the physical disk file operations by just using in-memory bytes instead, possibly improving performance and simplifying the code.

tests/across_compression.rs
```diff
@@ -144,6 +144,25 @@ fn compare_compression_contents_b44a_f16() {
     expect_eq_other("f16", "b44a.exr", "decompressed_b44a.exr");
 }
 
+// NOTE: these four DWA tests are expected to fail - not a bug in exrs.
```
johannesvollmer: I'll have to review this note more carefully and with more context

tests/dwa_csc.rs
```diff
@@ -0,0 +1,120 @@
...
+// bug - OpenEXR's own scalar and SIMD (SSE2/AVX) IDCT disagree with each
+// other (basis-constant precision and summation order both differ; see the
+// comments on `dct_inverse_8x8_scalar` in idct.rs), and real builds dispatch
+// to SIMD by default, so this port's default scalar-matching path differs
+// from a typical real decode by 1-2 ULP in half precision on some samples.
+// The `dwa_simd_identical` feature closes that gap (0 mismatches verified
+// against real decodes, including these two fixtures). This test's
+// tolerance is here to catch *structural* bugs (like the DC cursor bug
+// above), not to chase the inherent scalar/SIMD ambiguity.
+const MAX_ALLOWED_DIFF: f32 = 0.001;
```
johannesvollmer: We have a mechanism for approximate image equality checks, can't it be used here?

perhaps because this uses bin files? you could try to reuse the existing equality mechanism by using uncompressed exr as ground truth rather than bin files 

or perhaps don't even write files, do it in-memory using the rust bindings to the C++ OpenEXR library

Cargo.toml
```diff
@@ -49,6 +49,12 @@ default = ["rayon"]
 # rayon is used for parallel compression
 rayon = ["dep:rayon-core"]
 
+# Opt-in: unsafe x86_64 SSE2/AVX intrinsics that replicate
+# OpenEXR's SIMD-dispatched DWA decoder bit-for-bit (its basis constants and
+# summation order differ from OpenEXR's own scalar reference - see
+# src/compression/dwa/idct.rs). The only unsafe code in this crate; x86_64-only.
+dwa_simd_identical = []
```
johannesvollmer: I like the idea of being safe-only by default, but also allowing users to opt into unsafe optimizations. it makes sense as a general concept. perhaps it can be used in other parts of the library too. 

we could rename this to a generic `unsafe_optimizations` flag.

what I prefer even more would be safe simd though.

src/compression/dwa/idct_simd.rs
johannesvollmer, line 1: while I like the idea of feature-flagging unsafe optimizations, the one thing I would prefer even more is entirely safe simd code. there are rust libraries for this, or you can trust the compiler to do it by writing the code in a special way.

I would be happy to add a SIMD library to the project if it brings us a safe simd API (they can be unsafe inside). the `wide` crate seems promising on first glance. we should probably not use runtime detection because of the small data blocks then each having overhead.

please try to see if you can rewrite this file using a safe simd variant. Then we don't need the feature flag, and it can always be activated, providing performance speed ups for more users in practice. perhaps we can even remove the scalar variant then

src/compression/mod.rs
```diff
@@ -128,18 +129,15 @@ pub enum Compression {
...
     B44A, // TODO collapse with B44
 
-    /// __This lossy compression is not yet supported by this implementation.__
-    // lossy DCT based compression, in blocks
-    // of 32 scanlines. More efficient for partial buffer access.
-    DWAA(Option<f32>), /* TODO does this have a default value? make this non optional? default
-                        * Compression Level setting is 45.0 */
+    /// Lossy DCT-based compression (DreamWorks Animation), 32 scanlines per block.
+    /// Partial buffer access friendly.
+    /// Decoding support is implemented (encoding is still a stub).
```
johannesvollmer: it should not be a stub, it should return an error

src/compression/rle.rs
```diff
@@ -12,7 +12,22 @@ pub fn decompress_bytes(
...
+    interleave_byte_blocks(&mut decompressed_le);
+    super::convert_little_endian_to_current(decompressed_le, channels, rectangle)
+    // TODO no alloc
+}
+
+/// Classic byte-oriented run-length decoding shared by this compression method and DWA's RLE scheme (both port OpenEXR's `internal_rle_decompress`, see `compression::dwa`).
+/// A negative length-byte `-n` means "copy the next `n` bytes literally"; a non-negative `n` means "repeat the following byte `n + 1` times".
+/// Unlike full RLE decompression, this does not undo the delta prediction (`differences_to_samples`) or byte-block interleaving (`interleave_byte_blocks`) applied on top of it here, since DWA does not apply those extra steps.
+pub(super) fn unpack_rle_tokens(
+    compressed_le: &[u8],
```
johannesvollmer, line 27: nice

src/compression/rle.rs
```diff
@@ -12,7 +12,22 @@ pub fn decompress_bytes(
...
+    let mut decompressed_le = unpack_rle_tokens(&compressed_le, expected_byte_size, pedantic)?;
+    differences_to_samples(&mut decompressed_le);
+    interleave_byte_blocks(&mut decompressed_le);
+    super::convert_little_endian_to_current(decompressed_le, channels, rectangle)
+    // TODO no alloc
+}
+
+/// Classic byte-oriented run-length decoding shared by this compression method and DWA's RLE scheme (both port OpenEXR's `internal_rle_decompress`, see `compression::dwa`).
+/// A negative length-byte `-n` means "copy the next `n` bytes literally"; a non-negative `n` means "repeat the following byte `n + 1` times".
+/// Unlike full RLE decompression, this does not undo the delta prediction (`differences_to_samples`) or byte-block interleaving (`interleave_byte_blocks`) applied on top of it here, since DWA does not apply those extra steps.
```
johannesvollmer: remove the aspects of this comment which only paraphrase code. in general, the comments should prefer to answer "why" questions, not "what" questions

tests/dwa_csc.rs
```diff
@@ -0,0 +1,120 @@
...
+            mismatches,
+            actual.len(),
+            mismatch_fraction
+        );
+    }
+}
+
+#[test]
+fn dwa_three_standalone_lossy_dct_groups() {
+    check_against_real_openexr("y_ry_by_dwaa.exr", "y_ry_by_dwaa.bin", &["Y", "RY", "BY"]);
```
johannesvollmer: good approach. in the other tests we use uncompressed exr instead of bin files. that way, you can reuse the existing equality check from the library, and it will probably be easier to generate than an informal bin file. this also allows us to open them to visually verify that it matches. 

 but not a blocker for merging

johannesvollmer (commented):
thanks for your time! I appreciate the pull request. no need to worry about any previous commits or whatever.

I like your overall approach and I appreciate the tests you added. only the unsafe code is a difficult decision for me, since this was a unique selling point. I elaborate on this in the comments. 

I added a bunch of comments. feel free to discuss and brainstorm with me in the comments, instead of just starting with implementation right away, if you have concerns or questions.

johannesvollmer:
we already had (previously inactive) tests to compare DWAA images with visually equivalent png images in the repo. they are now executed automatically since you enabled DWAA compression.

unfortunately, it seems they fail, not just by precision loss, but drastical pixel differences (black instead of white). perhaps there is something wrong still?

not sure if the current GitHub pipeline triggers the SIMD path or the scalar path. would be nice to run all these tests for both paths in long term.

by the way I would also be ok with scalar-only implementation for the first version, especially since it is easier to test and to verify.

tests/images/valid/custom/dwa_csc/generate.py
zinedinp: [9244f982b539e2dfea2629c0dca3693286a5a369](https://github.com/johannesvollmer/exrs/commit/9244f982b539e2dfea2629c0dca3693286a5a369)

Yes, defenetly reduce dependencies and use OpenEXR and no bin files

I also used the already implemented lossy check with validate_result

src/compression/dwa/idct_simd.rs
zinedinp: I have looked inside wide and it seems promising, it doesnt looks too compicated but we may need to clarify target cpu (id use generic) and we need to use a higher minimum rust version, because https://crates.io/crates/wide/0.7.33 is the last version with 1.61. This isnt a problem iguess, but AVX512 isnt compatible and probably something unintetionally will break.

johannesvollmer: we could also do this in a follow up pr. this pr could focus on getting the scalar only version right, and optimization can come later

johannesvollmer: on second thought, perhaps it would be better to use runtime feature detection 

it seems `pulp` could perhaps do that safely. we should benchmark the performance though

src/compression/rle.rs
zinedinp: https://github.com/zinezockt/exrs/commit/394bee389ae05dce5c02d36712a18e0fb2ff6e97

src/compression/mod.rs
zinedinp: https://github.com/zinezockt/exrs/commit/c6f33a4cedae3bd5da58f0d520e47b700fde260e

Already produced an error, outdated comment

src/compression/dwa/idct_simd.rs
johannesvollmer: > I have looked inside wide and it seems promising, it doesnt looks too compicated but we may need to clarify target cpu (id use generic) and we need to use a higher minimum rust version, because https://crates.io/crates/wide/0.7.33 is the last version with 1.61. This isnt a problem iguess, but AVX512 isnt compatible and probably something unintetionally will break.

we could go with wide. since it depends on compile time flags, whoever compiles their project can determine which features are supported. but the fundamental problem remains that they usually don't know the CPU that will eventually run the code. therefore I think runtime feature detection would be better. I have not looked into pulp in detail yet

zinedinp: I made a rewrite of the scalar variant with wide, that produces the same values as the openexr reference coed.
But for the real deal i have prewritten an avx alterantive, maybe when feature building?

But i would wait until further instruction / code philosophy from you. (pulp / real bit identical / more specific code for other AVX512 (newer Version), RICSV)

zinedinp: > we could go with wide. since it depends on compile time flags, whoever compiles their project can determine which features are supported. but the fundamental problem remains that they usually don't know the CPU that will eventually run the code. therefore I think runtime feature detection would be better. I have not looked into pulp in detail yet

(pulp::Arch::new().dispatch(...) picks the best available implementation for the actual CPU), so we could make a hierarchie avx > sse2 > scalar

tests/across_compression.rs
johannesvollmer: Update: To make them pass, it should be possible to just set the one compression to the other, instead of setting both to uncompressed:

```diff
-            expected_decompressed.layer_data.encoding.compression = Compression::Uncompressed;
-            decompressed.layer_data.encoding.compression = Compression::Uncompressed;
+            expected_decompressed.layer_data.encoding.compression =
+               decompressed.layer_data.encoding.compression;
```

johannesvollmer: I ran the test and it passes then

johannesvollmer:
I just rudely pushed a fix for the tests not passing due to using exact equality into your branch :)

zinedinp:
I added pulp and revomed wide but lates pulp needed a higher rust version. Not ready to just merge.

 The test work on my AVX2 capeable CPU
 
expected to fail on Big-Endian Systems
 
 
 What to check:
 SSE2 CPU
 Scalar Fallback (ARM)

src/compression/dwa/idct_simd.rs
johannesvollmer: With pulp you write the alrogithm once. You write one algorithm that works on any simd width (scalar, 128, 256). pulp compiles a version of your algorithm for each possible width into the binary, and then calls the best one at runtime. we don't have to bother about selection or hierarchy. we only have to be able to write one width-independent algorithm. will this be possible for DWAA?

tests/across_compression.rs
johannesvollmer: I pushed this into your branch

src/compression/dwa/idct_simd.rs
zinedinp: > With pulp you write the alrogithm once.

This is true,that would be possible with"pulp::Simd + WithSimd + Arch::dispatch (or the #[pulp::with_simd] attribute macro)"
I didnt implement the scalar as a pulp function yet, so only V3 and V1 are actually pulp.

If we let pulp decide it, it wouldnt be tuned like the C kernel.
Should we future build or use the manuel kernel for V3 and V1 or something diffrent?


> and then calls the best one at runtime

currently my hierarchy is manuel:

`if let Some(v3) = V3::try_new() { avx::dct_inverse_8x8(v3, data); return; }
if let Some(v1) = V1::try_new() { sse2::dct_inverse_8x8(v1, data); return; }
dct_inverse_8x8_scalar(data);`

But if we let pulp decide, it would do the same hierarchy (With more SIMD that are not in the OpenEXR, shouldnt be a huge Problem tho)


> will this be possible for DWAA?

This is defentely possible. We could Write the pulp ground algorithm like the avx or the see2 kernel, so one would match.

src/compression/dwa/idct.rs
johannesvollmer, line 1: this approach seems good overall. just the testing is the last open question.

src/compression/dwa/idct.rs
```diff
@@ -0,0 +1,645 @@
...
+pub fn dct_inverse_8x8_dc_only(data: &mut [f32; 64]) {
+    let val = data[0] * 0.3535536f32 * 0.3535536f32;
+    for v in data.iter_mut() {
+        *v = val;
+    }
+}
+
+// All tests exercise the SIMD kernels; only on x86.
+#[cfg(all(test, any(target_arch = "x86", target_arch = "x86_64")))]
+mod tests {
```
johannesvollmer: good idea for the tests, but i really would not like these tests to silently fail sometimes. i don't know a solution right now. perhaps we can use the same mechanism in github workflows which emulates big endian system to emulate avx256 support? otherwise, how do we make sure it works?

tests/dwa_csc.rs
```diff
@@ -0,0 +1,77 @@
...
+    // Match the ground truth's lossless encoding so "validate_result"
+    // compares the samples bit-for-bit instead of with a lossy tolerance.
+    decoded.layer_data.encoding = ground_truth.layer_data.encoding;
+
+    ground_truth.assert_equals_result(&decoded);
+}
+
+#[test]
+fn dwa_three_standalone_lossy_dct_groups() {
+    check_against_real_openexr("y_ry_by_dwaa.exr", "y_ry_by_dwaa_ground_truth.exr");
```
johannesvollmer: nice!

tests/images/valid/custom/dwa_csc/generate.py
```diff
@@ -0,0 +1,91 @@
...
+    ground_truth_header['channels'] = {name: Imath.Channel(HALF) for name in channel_names}
+    ground_truth_header['compression'] = Imath.Compression(Imath.Compression.NO_COMPRESSION)
+
+    ground_truth_out = OpenEXR.OutputFile(path_ground_truth_exr, ground_truth_header)
+    ground_truth_out.writePixels(ground_truth_pixels)
+    ground_truth_out.close()
+
+
+write_and_dump(
+    "y_ry_by_dwaa.exr", "y_ry_by_dwaa_ground_truth.exr",
```
johannesvollmer, line 80: nice!

Cargo.toml
```diff
@@ -12,7 +12,9 @@ repository = "https://github.com/johannesvollmer/exrs"
 readme = "README.md"
 license = "BSD-3-Clause"
 exclude = [ "specification/*", "specification/**", "tests/images/*", "tests/images/**" ]
-rust-version = "1.61.0"
+# 1.83 is required by pulp (verified by compiling: 1.82 fails on
+# raw-address syntax and more, 1.83 builds the whole crate)
+rust-version = "1.83.0"
```
johannesvollmer, line 15: rust-version bump is ok, since our main downstream dependant (`image-rs`) is already at `1.88` 👍

Cargo.toml
```diff
@@ -34,11 +36,13 @@ miniz_oxide = "^0.8.0"         # zip compression for pxr24
 smallvec = "^1.7.0"            # make cache-friendly allocations        TODO profile if smallvec is really an improvement!
 rayon-core = { version = "^1.11.0", optional = true }         # threading for parallel compression
 zune-inflate = { version = "^0.2.3", default-features = false, features = ["zlib"] }  # zip decompression, faster than miniz_oxide
+pulp = "0.22.3"                # safe *runtime* SIMD dispatch (avx2 > sse2 > scalar) for the DWA decoder's inverse DCT, mirroring OpenEXR's own cpuid dispatch. Needs default features: "std" for is_x86_feature_detected!, "x86-v3" for the AVX2 token type.
 
 [dev-dependencies]
 image = { version = "0.25.2", default-features = false, features = ["png"] }         # used to convert one exr to some pngs
 
 bencher = "0.1.5"
+pulp = "0.22.3"           # tests check the CPU's SIMD tier to know which decodes must be bit-exact (see tests/dwa_csc.rs)
```
johannesvollmer: can't you use existing dev dependencies in tests? since we already added it above. if this needs to stay, replace the comment with an explanation why this has to appear twice

Cargo.toml
```diff
@@ -34,11 +36,13 @@ miniz_oxide = "^0.8.0"         # zip compression for pxr24
 smallvec = "^1.7.0"            # make cache-friendly allocations        TODO profile if smallvec is really an improvement!
 rayon-core = { version = "^1.11.0", optional = true }         # threading for parallel compression
 zune-inflate = { version = "^0.2.3", default-features = false, features = ["zlib"] }  # zip decompression, faster than miniz_oxide
+pulp = "0.22.3"                # safe *runtime* SIMD dispatch (avx2 > sse2 > scalar) for the DWA decoder's inverse DCT, mirroring OpenEXR's own cpuid dispatch. Needs default features: "std" for is_x86_feature_detected!, "x86-v3" for the AVX2 token type.
```
johannesvollmer: shorten the comment. just "used for runtime SIMD dispatch" or similar

README.md
```diff
@@ -57,7 +57,7 @@ __What we can do:__
             - [x] little-endian architectures
             - [ ] big-endian architectures __(help wanted)__
         - [x] b44, b44a (huge thanks to @narann)
-        - [ ] dwaa, dwab __(help wanted)__
+        - [~] dwaa, dwab __(help wanted)__
```
johannesvollmer: add two sub points, one checked for decoding, one unchecked for encoding

tests/dwa_csc.rs
```diff
@@ -0,0 +1,77 @@
...
+fn dir() -> &'static Path {
+    Path::new("tests/images/valid/custom/dwa_csc")
+}
+
+// The ground truth was produced by an OpenEXR build that picked its avx
+// IDCT kernel. This code dispatches to the same kernel hierarchy at runtime
+// (src/compression/dwa/idct.rs), so on an AVX2-capable machine the decode
+// must be bit-identical. On lower tiers exrs correctly picks a different
+// kernel (sse2/scalar) whose output legitimately differs from the avx-made
+// ground truth by a few ULP, so there is nothing meaningful to compare and
```
johannesvollmer: i disagree, even if it is not exactly the same, we should test that it is roughly the same

johannesvollmer (commented):
looks good overall now 🚀  there is just one last open question, the tests.

### first, let me explain what the problem is:

- the current tests silently skip when the machine does not support AVX256
- the github workflow runner machine likely does not support AVX256, a breakage in the future would go unnoticed easily
- there is no easy way to check for me whether the tests were run successfully, or whether they were skipped completely (i don't know if i have AVX256)
- there is no easy way to tell which one of the SIMD implementations were run

### i think i have a solution:
- use some kind of emulation which devs can run locally and the workflow can also run, probably the Intel SDE (Software Development Emulator)
- add a rust flag called `avx256-tests` and another called `sse128-tests` 
- when the flag is used, no tests except the SIMD based tests should be run (to avoid extremely long test times). this can be achieved by moving all these tests to a separate integration file in `tests/simd.rs` and compiling the project with `cargo test --test avx2`
- the workflow and devs can run the test binary using SDE emulation to check the SIMD correctness

I will research this, but you can also dive into it, if you want

johannesvollmer:
Robot said the github workflow could look similar to this:

```yaml
name: SIMD tests

on:
  push:
  pull_request:

jobs:
  avx2-sde:
    runs-on: ubuntu-latest
    timeout-minutes: 90

    env:
      SDE_VERSION: sde-external-10.8.0-2026-03-15-lin
      SDE_SHA256: 50B320CD226ACEF7A491F5B321FC1BE3C3C7984F9E27A456E64894B5B0979DD3
      SDE_URL: https://downloadmirror.intel.com/859732/sde-external-10.8.0-2026-03-15-lin.tar.xz

    steps:
      - uses: actions/checkout@v4

      - uses: dtolnay/rust-toolchain@stable

      - uses: Swatinem/rust-cache@v2

      - name: Cache Intel SDE
        id: cache-sde
        uses: actions/cache@v4
        with:
          path: ~/.cache/intel-sde/${{ env.SDE_VERSION }}
          key: intel-sde-${{ env.SDE_VERSION }}

      - name: Download Intel SDE
        if: steps.cache-sde.outputs.cache-hit != 'true'
        run: |
          mkdir -p ~/.cache/intel-sde
          curl -L "$SDE_URL" -o /tmp/sde.tar.xz
          echo "$SDE_SHA256  /tmp/sde.tar.xz" | sha256sum -c -
          mkdir -p ~/.cache/intel-sde/"$SDE_VERSION"
          tar -xf /tmp/sde.tar.xz --strip-components=1 -C ~/.cache/intel-sde/"$SDE_VERSION"

      - name: Check SDE
        run: |
          ~/.cache/intel-sde/"$SDE_VERSION"/sde64 --version

      - name: Build AVX2 integration test only
        run: |
          RUSTFLAGS="-C target-feature=+avx2,+fma" \
          cargo test --test avx2 --features avx2-tests --no-run --message-format=json \
            | jq -r 'select(.reason == "compiler-artifact" and .target.name == "avx2" and .executable != null) | .executable' \
            | tail -n 1 > /tmp/avx2-test-bin

          test -s /tmp/avx2-test-bin
          cat /tmp/avx2-test-bin

      - name: Run AVX2 integration test under Intel SDE
        run: |
          TEST_BIN="$(cat /tmp/avx2-test-bin)"
          ~/.cache/intel-sde/"$SDE_VERSION"/sde64 -hsw -- "$TEST_BIN" --nocapture
```

The exact feature flags and commands may have to be adjusted to make sure they match this repo and respect runtime dispatch.

johannesvollmer:
Update: We coud perhaps use `cross` which we already use, for the SDE emulation. This would be awesome! 

I'm working on it over at https://github.com/johannesvollmer/exrs/tree/DWA-decomp-with-github-workflow to test the ci without disturbing your PR. When it works, you can pull those changes into your PR.

johannesvollmer:
Update: 
image

AI got AVX2 pipeline working

johannesvollmer:
Both are working now. See this branch and pull the required files from there:
https://github.com/johannesvollmer/exrs/pull/274

The test also had to be moved into the /tests directory for a better test suite selection command. The README and workflow file are especially important.

I just realized we forgot the another crucial operation: We should **measure** whether these SIMD implementations are actually faster than the scalar ones. The most important step before adding optimizations. We should add a benchmark, forcing a specific path (either scalar or ss2 or avx) and run it on non-emulator hardware. Let me try something...

There is a good reason to question the speed of the SIMD implementation: Currently, the CPU architecture is checked for each invocation of the `dct_inverse` function - this might be too much overhead and might result in worse performance.

zinedinp:
> We should measure whether these SIMD implementations are actually faster than the scalar ones

What a crucial part that I totally missed. 
Pulp's vectorize doc comment says explicitly: https://github.com/sarah-quinones/pulp/blob/main/pulp/src/core_arch/mod.rs#L357-L362

And A Batch version for avx2 was added to not call the function everytime

And I made some other corrections

README.md
johannesvollmer: I can also do it, we can merge without this, don't worry about it

johannesvollmer:
Nice! Great work! Every check passes. I had another detailed review. Test coverage looks good and I am confident in the code.

I just ran the benchmarks though:

    test bench_avx2       ... bench:     123,050 ns/iter (+/- 12,465)
    test bench_avx2_batch ... bench:     122,664 ns/iter (+/- 14,314)
    test bench_scalar     ... bench:     126,012 ns/iter (+/- 12,589)
    test bench_sse2       ... bench:     121,540 ns/iter (+/- 13,835)

I must say the gains seem to be quite modest on my machine. Around 1% improvement it seems. Could be noise. Are these the same numbers you also got?

One thing to note is that Rust on x86 has SSE2 always enabled, therefore, even the scalar function got secretly vectorized to SSE2, so there is never a truly scalar version to begin with. Therefore, the bench only compares SSE2 with AVX, not with scalar. The only case where the scalar branch would actually be executed, is on non-x86 machines, such as ARM (many mobile devices and a few laptops).

~Thinking further: This indicates that we could have simply used pulp's `auto_vectorize` function and not hand-roll the sse2 version at all. It would still give us the exact same behaviour on all platforms, but with a only third of the code needed.~

~I know it's hard to delete code that you spent hours to get right. But since each line of code has a cost in project maintenance, we should really weigh our options. What do you think? Or perhaps, am I missing something?~

Edit: What we get with manual vectorization is the guarantee that it is vectorized, where the compiler does not guarantee successfull auto vectorization. So that's it, then let's merge. Any last words before merging?

johannesvollmer (approved):
Great work. Thanks for your continuous efforts! 🏅

zinedinp:
Great to hear that you are happy with the work.

```
test bench_avx2       ... bench:      66,193 ns/iter (+/- 8,978)
test bench_avx2_batch ... bench:      68,026 ns/iter (+/- 8,336)
test bench_scalar     ... bench:      86,050 ns/iter (+/- 8,925)
test bench_sse2       ... bench:      76,201 ns/iter (+/- 8,458)
```

perheaps i get much clearer results. (R7 9800X3D)

>  One thing to note is that Rust on x86 has SSE2 always enabled, therefore, even the scalar function got secretly vectorized to SSE2, so there is never a truly scalar version to begin with

Thats true, all cpu since the early 2000 have sse2, so the scalar would never be used on any CPU in use today.
I would leave it there as the reference calculation and make if case for when the programm that uses this crate in x86 only, or only compiles for x86 that the scalar isn't compiled.

Secound option is to build the sse2 algorithm as ground truth, so we would have only an additional avx2 and ARM CPUs would get a sse2 solution, but this would diverge from OpenEXR.

I would put one of these things in the todo list, but I am satisfied with the code, lets get this bad boy a new version number.

johannesvollmer:
I think the scalar will be used on non-x68, which means ARM mobile devices and a few laptops, and perhaps older wasm. Let's keep it like this 💯

johannesvollmer:
Nice to see it having more impact on other CPUs. Seems unwise of me to only consider my own CPU.

johannesvollmer:
Thanks again for contributing this long-awaited feature! It has been a please to work with you.

PR #279, opened 2026-07-14
Large Huffmann rewrite for DWA and Cache Size optimizeation

zinedinp:
Good evening,
I’ve been doing a bit of testing to find out why DWA was still slower than the OpenEXR implementation, and I’ve noticed that Huffman decoding in the AC stream takes much longer than in OpenEXR.
Over the past few days, Ive implemented the ideas from https://ieeexplore.ieee.org/document/11510463 .
Some of these ideas are already incorporated into OpenEXR’s `internal_huffman`, but others are not.

There are still some areas for improvement, which I’ll get round to when I can.

But I’d like you to have a look at this and see if we want to proceed with improving the Huffman decoding.
Should we keep the old Huffman decoder as a fallback, or go for a universally optimised, non-textbook version?

I suggest reading the extensive header comment in huffman.rs


johannesvollmer:
Hi! Performance improvements are welcome of course. I performed a few refactorings after you last pr, mainly splitting up code into smaller files. Unfortunately this will result in merge conflicts if you didn't pull those changes yet, sorry!

zinedinp:
No problem i helped myself with AI (looks good, it copied fine).

It is not ready for a merge, because the huffman also touches piz and i dint test that.

Also some impovements are still open.
I am open for inputs and the question if we want a fallback or only a optimized version.

zinedinp:
Huffman-Decoder huffman.rs)build to DCC-2026-Verfahren (Zavadskyi & Kovalchuk)

Around 90 to 100% speedgain for the huffman decode, also works for piz.

Next step would be:
a Huffman-Symbol-Loop
Tablebuild without 65537
Huffman SIMD implementation (biggest gain)

zinedinp:
I combined all my ideas from the last weak and its more then just the huffman.

I feel confident, that a push is possible and we gain around 10% overall.

You might want to take a look how the L3 cache limit was solved to not read off memory

zinedinp:
@johannesvollmer would be nice if merged in near future

zinedinp:
making a larger then expected fork, new pull will come sometime

